Issue #103, opened 2015-05-23
Filter unwanted elements from wollok projects in Package View


javierfernandes:
We are currently using the Package Explorer view for showing wollok projects in our perspective
But this adds a lot of "noise" from java projects that we don't want the user/student to be aware of of.

![screen shot 2015-05-23 at 20 45 36](https://cloud.githubusercontent.com/assets/4428120/7786108/d851c23a-018c-11e5-9fc0-6ddf5dc7a25e.png)

Check out this example: theres libraries, JRE,  META-INF, build.properties, etc.

We would like something cleaner.

javierfernandes:
Done !

I've discovered that there's an extension point to add custom filters for java elements.
I've created one that applies to the Package Explorer view.

Now it shows wollok projects like this:

![screen shot 2015-05-24 at 10 16 14](https://cloud.githubusercontent.com/assets/4428120/7787979/6f7057f0-01fe-11e5-8cb9-7412cbd90a7f.png)

PR #104, opened 2015-05-24
#103 Filter unwanted java elements from Projects view (Package Explorer)

javierfernandes:
See for screenshot #103


javierfernandes:
I made a mistake. I wanted to create it against "dev" and not master.
I don't know how to change that. So I'll create a new one

PR #376, opened 2015-11-11
100% natives objects now coded in wollok as part of the SDK

javierfernandes:
SDK now includes: Collection, List, Set, Object, Integer, Number, Double, Range, Boolean, String, and InstanceVariableMirror.

They have minor behaviour, because still the methods are native, but this native impl is consistent with the users level native objects.

Now the interpreter doesn't have any special treatment for this objects


javierfernandes:
heavy PR @tesonep @npasserini

org.uqbar.project.wollok.launch/src/org/uqbar/project/wollok/debugger/server/rmi/XDebugStackFrameVariable.xtend
```diff
@@ -21,7 +20,8 @@ class XDebugStackFrameVariable implements Serializable {
 	}
 	
 	def dispatch asRemoteValue(WollokObject object) { new XWollokObjectDebugValue(variable.name, object) }
-	def dispatch asRemoteValue(WollokList list) { new XWollokListDebugValue(list) }
+	// TODO: after changing lists to wollokobjects
+//	def dispatch asRemoteValue(WollokList list) { new XWollokListDebugValue(list) }
 	def dispatch asRemoteValue(Object o) { new XDebugValue(o.toString) }
```
npasserini: Is this still necessary?

javierfernandes: Yeap, good that you noticed it. I forgot about that. I broke the special visualization of lists in the debugger UI.
I need to fix that. But I guess that it could be after merging

javierfernandes:
It is highly probably that with this change I have broken something in the collections behavior, so we need to test it manually and increase junit tests.

Here is a potential problem that I worked around. The "detect" method was implemented like this

```xtend
		method detect(predicate) = this.fold(null, [ acc, e | if (acc == null && predicate.apply(e)) e else acc ])
```

But tests failed, things like returning the last object when it should return null because none were supposed to match, or returning the previous call value (really weird).
Refactoring to this worked fine

```xtend
		method detect(predicate) = this.fold(null, [ acc, e |
			 if (acc != null)
			 	acc
			 else
			 	if (predicate.apply(e)) e else null
		])
```

Let me know if you have an idea on what could be the cause. I was not able to find any problem looking at this code. Seems right

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -21,16 +21,46 @@ package lang {
 	 * @author jfernandes
 	 * since 1.0
 	 */
-	class WObject {
+	class Object {
 		method identity() native
+		method instanceVariables() native
```
npasserini: Why do we need this?
I would prefer that wollok objects do not understand this kind of reflective methods and use a mirror instead.

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -21,16 +21,46 @@ package lang {
...
+		method equals(other) = this == other
+		
 		method ->(other) {
 			return new Pair(this, other)
 		}
+		
+		method randomBetween(start, end) native
+		
+		method toString() {
+			// TODO: should be a set
```
npasserini: Why is it not a set?

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -45,7 +75,249 @@ package lang {
...
+					if (criteria.apply(n, acc.getY()))
+						new Pair(e, n)
+					else
+						acc
+				}
+			])
+			return if (result == null) null else result.getX()
+		}
+		 
+		// non-native methods
```
npasserini: I think this comment is wrong because there are more non-native methods before it

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -45,7 +75,249 @@ package lang {
...
+	 */
+	class List extends Collection {
+
+		method get(index) native
+		
+		override method newInstance() = new List()
+		
+		method any() {
+			if (this.isEmpty()) 
+				throw new Exception() //("Illegal operation 'any' on empty collection")
```
npasserini: Why not 
throw new Exception("Illegal operation 'any' on empty collection")
?

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -45,7 +75,249 @@ package lang {
...
+		method <=(other) native
+		
+		method abs() native
+		method invert() native
+	}
+	
+	/**
+	 * @author jfernandes
+	 * @since 1.3
+	 * @noInstantiate
```
npasserini: Why is this annotation only in Integer/Number and not in other primitive types (e.g. Boolean)

org.uqbar.project.wollok.lib/src/wollok/lang/WSet.xtend
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * @author jfernandes
+ */
+class WSet extends WCollection<Set> {
+	
+	new() {
+		wrapped = newHashSet
+	}
+	
+	def first() { wrapped.head }
```
npasserini: Sets should not implement 'first'

org.uqbar.project.wollok.lib/src/wollok/lib/AssertObject.xtend
```diff
@@ -27,13 +30,13 @@ class AssertObject extends AbstractWollokDeclarativeNativeObject {
...
 	def assertEquals(Object a, Object b) {
-		if (asBinaryOperation("==").apply(a, [|b]) == false)
+		if (asBinaryOperation("==").apply(a, [|b]).wollokToJava(Boolean) == false)
 			throw AssertionException.valueNotWasEquals(a, b)
 	}
 
 	@NativeMessage("notEquals")
 	def assertNotEquals(Object a, Object b) {
-		if (asBinaryOperation("==").apply(a, [|b]) == true)
+		if (asBinaryOperation("==").apply(a, [|b]).wollokToJava(Boolean) == true)
```
npasserini: "==true" is not necessary

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NumberTestCase.xtend
```diff
@@ -10,6 +10,22 @@ import org.junit.Test
...
+	def void integersAsResultValueOfNative() {
+		'''program a {
+			assert.equals(4, "hola".length())
+		}
+		'''.interpretPropagatingErrors
+	}
+	
+	@Test
+	def void sum() {
+		'''program a {
```
npasserini: I think it would be better to create tests instead of programs

org.uqbar.project.wollok/src/org/uqbar/project/wollok/interpreter/WollokInterpreterEvaluator.xtend
```diff
@@ -123,13 +125,16 @@ class WollokInterpreterEvaluator implements XInterpreterEvaluator {
...
 	def dispatch Object evaluate(WIfExpression it) {
 		val cond = condition.eval
+		
 		// I18N !
-		if (!(cond instanceof Boolean)) throw new WollokInterpreterException(
-			"Expression in 'if' must evaluate to a boolean. Instead got: " + cond, it)
-		if (Boolean.TRUE == cond)
+		if (!(cond.isWBoolean)) 
+			throw new WollokInterpreterException('''Expression in 'if' must evaluate to a boolean. Instead got: «cond» («cond?.class.name»)''', it)
+		if (wollokToJava(cond, Boolean) == Boolean.TRUE)
```
npasserini: I think == Boolean.TRUE shoudn't be necessary

org.uqbar.project.wollok/src/org/uqbar/project/wollok/interpreter/api/WollokInterpreterAccess.xtend
```diff
@@ -25,19 +26,7 @@ class WollokInterpreterAccess {
 	 * which is needed in different parts of the interpreter 
 	 */
 	def boolean wollokGreaterThan(Object a, Object b) {
-		operations.asBinaryOperation(">").apply(a, [|b]).isTrue()
+		operations.asBinaryOperation(">").apply(a, [|b]).isTrue
```
npasserini: I am not sure that > should be a lazy operation.

javierfernandes: It's not. Is just that now the contract of "asBinaryOperation" has been broaden so the function that it returns always a closure as second param. But most operations just return the same "b" object which is b already evaluated.
The only lazy ops are && and ||. The actual code that has makes it lazy is the calling method in the evaluator (similar to this one) but that is something like this
```
operations.asBinaryOperation("&&").apply(a, [| b.eval() ])
```

npasserini:
Yeah, it looks right to me too!

org.uqbar.project.wollok/src/org/uqbar/project/wollok/interpreter/operation/WollokDeclarativeNativeBasicOperations.xtend
```diff
@@ -61,12 +66,7 @@ class WollokDeclarativeNativeBasicOperations implements WollokBasicBinaryOperati
 		else if (b == null) a
 		else sum(a, b)
 	}  
-		def dispatch sum(Object a, Object b) { throw new IllegalBinaryOperation('''Unable to add «a» + «b»''') }
-		def dispatch sum(WollokInteger a, WollokInteger b) { new WollokInteger(a.wrapped + b.wrapped) }
-		def dispatch sum(WollokNumber<?> a, WollokNumber<?> b) { new WollokDouble(a.doubleValue + b.doubleValue) }
-		def dispatch sum(WollokNumber<?> a, Object b) { throw new IllegalBinaryOperation('''Unable to perform «a» («a.class?.simpleName») + «b» («b.class?.simpleName»)''') }
-		def dispatch sum(String a, Object b) { a + b }
-		//
```
npasserini: Nice

npasserini:
:+1:

Issue #403, opened 2015-11-14
Replace extend keyword with subclass 


flbulgarelli:
Although Java derived languages tend to use `extend` keyword for subclassing, it is a misleading word. 

We don't use the term `extend` (extender) in our speech in initial OO courses. Instead, we use inherit, subclass, (heredar, subclasificar), since we usually reserve the meaning of extension to the fact of adding behavior to an existing object or class - using extension methods or open classes, for instance. 

So we have here an inconsistency between the language syntax and our speech. My suggestion here is to rename the `extend` keyword with `subclass` - or perhaps `inherit`

npasserini:
I am not sure about subclass, it seems weird for me:

class Golondrina subclass Ave {}

Maybe we should write
class Golondrina subclass of Ave {}

Still does not fit too much

Inhterits feels a little better,

class Golondrina inherits Ave {}
or
class Golondrina inherits from Ave {}

Please note the final s of inherits. This is because I read it as a
declaration (a golondrina is a kind of ave) and not as an order (create a
class inheriting from ave). I prefer the declarative semantics and, well,
the keyword should reflect it.

javierfernandes:
This is already merged into "dev". I guess we can already mark it as solved. It will be released in Wollok 1.3.0

javierfernandes:
Closing it.
To be released with v1.3.0

PR #413, opened 2015-11-15
Issue #403 replace extends keyword

flbulgarelli:


javierfernandes:
We need to update the Language Reference.
The only issue is that now the syntax is different depending on the version.
So, probably we need to start having duplicated wiki pages for each version (this one will be 1.3). At least for the language reference.

Can you please do that ? I'll add this to 1.3 release notes (also in the wiki)

flbulgarelli:
@javierfernandes  no problem. I'll do that. 

Just I need to edit the wiki or something else?

PR #416, opened 2015-11-15
Issue #414 replace this with self

### LINKED ISSUE #414: Replace this with self
C++ derived languages - JS, Java, C# - tend to use `this` keyword for object self-referencing, while Python and Smalltalk derived languages - swift, self, ruby - tend to use `self`. 

Thus both  `self` and `this` keywords are _common_ terms among programmers. 

However, using `self` has three good points over `this`: 
   * backward compatibility with our speech: dozens of exams and didactic texts we have written before use the term `self`. Keeping it saves migration energy. 
   * conceptual clarity: `self` better reflects the idea of **self** referencing. Perhaps those terms are very similar in the context of an object declaration, but not a class declaration:  since classes are not object in wollok, `self` (si mismo)  can hardly mentally be mapped to the class. But `this` (esto) sounds more lexically scoped: this seems to be a reference to the class being declarated instead of the object. 
   * object point of view: `self` enforces the idea that objects make runtime decisions, instead of the view in which the programmer manipulates them and makes decisions form them. Although Wollok is not a pure live-object-system like Smalltalk, the idea of live objects in runtime, that collaborate among themselves is powerful and declarative. `this` keyword, on the other hand, has a more imperative meaning: _when this happens, do that on **this** thing_. Its very subtle, but subtleness is essential.

flbulgarelli:
Fixes #414

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -25,11 +25,11 @@ package lang {
 		method identity() native
 		
 		method ==(other) {
-			return this === other
+			return self === other
 		}
 		
 		method ->(other) {
-			return new Pair(this, other)
+			return new Pair(self, other)
 		}
 	}
 	
@@ -42,8 +42,8 @@ package lang {
 		}
 		method getX() { return x }
 		method getY() { return y }
-		method getKey() { this.getX() }
-		method getValue() { this.getY() }
+		method getKey() { self.getX() }
+		method getValue() { self.getY() }
 	}
  
 }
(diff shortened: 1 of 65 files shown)


flbulgarelli:
Finally we decided to merge this PR, it will be part of 1.4. I am going to resolve conflicts and merge it.

flbulgarelli:
@fdodino @npasserini @matifreyre

flbulgarelli:
Replaced by #611

javierfernandes:
Can we delete this branch ? Has 1 unmerged commited.
I'm cleaning up branches

npasserini:
I think yes.

Issue #418, opened 2015-11-15
Create a new tool to print a given wollok file's AST


javierfernandes:
Print output in JSON format.

javierfernandes:
For this example input file:

```xtend
class Asustador {
	var edad = 0
	var nivelMotivacion = 100
	
	new(e) { edad = e }
	method getEdad() = edad  
	method setEdad(e) { edad = e }
	
	method getNivelMotivacion() = nivelMotivacion
	
	method entrarAPuerta(puerta) {
		return puerta.entra(this)
	}
	
	method asustar(ninio) {
		val a = this.getPorcentaje()
		return a * this.puntosDeTerror() / ninio.getEdad()
	}
	method getPorcentaje() {
		return nivelMotivacion / 100
	}
	
	method puntosDeTerror()
	
	method reducirMotivacion(cuantoPorciento) {
		nivelMotivacion -= cuantoPorciento * nivelMotivacion
	}
}

class AsustadorNato extends Asustador {
	var puntosTerrorInnatos
	new(e, p) = super(e) { 
		puntosTerrorInnatos = p
	}
	override method puntosDeTerror() {
		return puntosTerrorInnatos * this.getEdad()
	}
}

class AsustadorPerseverante extends Asustador {
	var puntosDeTerror = 0
	
	new(e) = super(e)
	
	method mejora(actividad) {
		puntosDeTerror += actividad.calcularMejora()
	}
	override method puntosDeTerror() = puntosDeTerror
}

```
It generates the following JSON

```json
{
		"type" : "File",
		"elements" : [ {
				"type" : "Class",
				"name" : "Asustador",
				"members" : [ {
						"type" : "Variable",
						"variable" :  {
								"type" : "Variable",
								"name" : "edad"
							}, 
						"right" :  {
								"type" : "Number",
								"value" : "0"
							} 
					},{
						"type" : "Variable",
						"variable" :  {
								"type" : "Variable",
								"name" : "nivelMotivacion"
							}, 
						"right" :  {
								"type" : "Number",
								"value" : "100"
							} 
					},{
						"type" : "Method",
						"name" : "getEdad",
						"expression" :  {
								"type" : "VariableReference"
							} 
					},{
						"type" : "Method",
						"name" : "setEdad",
						"parameters" : [ {
								"type" : "Parameter",
								"name" : "e"
							} ],
						"expression" :  {
								"type" : "Block",
								"expressions" : [ {
										"type" : "Assignment",
										"feature" :  {
[...]

PR #419, opened 2015-11-15
Dev astprinter

javierfernandes:
Take a look and see if it is enough for Mumuki

### DIFF
--- a/org.uqbar.project.wollok.launch/src/org/uqbar/project/wollok/launch/WollokASTAnalyzer.xtend
+++ b/org.uqbar.project.wollok.launch/src/org/uqbar/project/wollok/launch/WollokASTAnalyzer.xtend
@@ -2,51 +2,67 @@ package org.uqbar.project.wollok.launch
 
 import com.google.inject.Injector
 import java.io.File
+import java.lang.reflect.Method
+import java.util.List
 import org.eclipse.emf.ecore.EObject
 import org.eclipse.emf.ecore.resource.Resource
 import org.uqbar.project.wollok.wollokDsl.WFile
-import java.util.List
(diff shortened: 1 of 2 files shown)


javierfernandes:
See comments in #418

org.uqbar.project.wollok.launch/src/org/uqbar/project/wollok/launch/WollokASTAnalyzer.xtend
```diff
@@ -0,0 +1,68 @@
...
+	
+	def features(EObject it) { eContents.groupBy[eContainingFeature].entrySet.map[key?.name -> elements(value, key)] }
+	
+	def elements(List<EObject> it, EStructuralFeature feature) '''«IF feature.many»[«ENDIF» «map[e|process(e)].join(",")» «IF feature.many»]«ENDIF»'''
+	
+	// *************************************
+	// ** names transformations
+	// *************************************
+	
+	val transformations = #[ [removePrefix("W")], [removeSuffix("Expression")], [removeSuffix("Declaration")], [removeSuffix("Literal")] ]
```
flbulgarelli: I would keep AST elements as-is. This makes the mapping between the grammar and the generates AST obvious.

flbulgarelli: OK, I dint't see you comments in the issue. Now I understand that you are going to rename those elements in the near future.

flbulgarelli:
@faloi Look, this is awesome :) It is based on http://esprima.org/, so it should be easy to process like we did in the hspec/mocha runners. 

@AlvarezAriel  could we have something like this implements for the new Gobstones?

Issue #429, opened 2015-11-17
Model MessageNotUnderstood exception in wollok itself


javierfernandes:
We need to model it as a regular wollok Exception

npasserini:
Why?

javierfernandes:
Ok, maybe it was not 100% necessary :P  But it will simplify the interpreter to have it modeled as another regular wollok exception, and reuse the same code that shows an exception.
I guess it is "elegant" in the sense that the interpreter should be easier.

I'm not doing it to have a dynamic programming model. I'll prefer to have a simpler type system and not dynamic facilities.
I guess that we can limit this method to avoid dynamic objects and therefore make the type system easier.

javierfernandes:
Sample outputs:

```
wollok.lang.MessageNotUnderstoodException: a AsustadorNato[nivelMotivacion=100, edad=25, puntosTerrorInnatos=200] does not understand cantina()
	at wollok.lang.Object.messageNotUnderstood(name,parameters) [/wollok.wlk:151]
	at  [/workspace.wpgm:23]
```

<img width="489" alt="screen shot 2015-11-17 at 00 22 29" src="https://cloud.githubusercontent.com/assets/4428120/11216423/db350466-8d29-11e5-8fdb-7b12ff68ba8c.png">
<img width="486" alt="screen shot 2015-11-17 at 00 33 01" src="https://cloud.githubusercontent.com/assets/4428120/11216422/db310b2c-8d29-11e5-931a-cd501c78a4a3.png">

PR #432, opened 2015-11-17
Dev #429 message not understood in wollok #434 json format for test output

javierfernandes:
This PR has 3 changes
* better exception messages for any wollok.lang.Exception. We don't print the internal java stack but the wollok stack. 
* modelling the MessageNotUnderstood as a wollok.lang.Exception so that it is handled in the same way as any other exception. 
* additionally, the MessageNotUnderstood is being thrown by each objects with a doesNotUnderstand() message. This might be something we could take back in the future, since we don't want dynamic objects

Also some fixes in the wollok sdk library like it Object was missing the "!=" method. And adding "map" method to ranges.

I had to change the way the launcher talks to the UI for wollok tests, because to show a Wollok exception it needs to evaluate the interpreter by sending the message getStackTrace() to the wollok object (exception). And we don't want to serialize all the object's graph along with the interpreter to the UI for that.
So the wollok process is translating from the exception into a string, and sending that string through the wire to the UI.

-----------------

Besides all of this infraestructure for better error handling I had to start a new feature on top of this, so it is in this same PR. It's the JSON output for tests. This will make possible the mumuki integration.


javierfernandes:
@tesonep  could also help reviewing the changes to the test part :)

Issue #369, opened 2015-11-02
Introduce wollok.lang.Closure


javierfernandes:
I believe that this requires either having many classes like Closure0, Closure1, Closure2, Closure3.. .etc.. for different numbers of parameters, either supporting varargs for the "apply method"

```
method apply(arguments...) {
}

```

npasserini:
I like the varargs!

javierfernandes:
Varargs already implemented in #371 
Now we just need to model the wollok.lang.Closure and remove the 100% native class and refactor the whole interpreter

javierfernandes:
Done !

For the moment, it just has this

```xtend
	/**
	 * @author jfernandes
	 * @since 1.3
	 * @noInstantiate
	 */
	class Closure {
		method apply(parameters...) native
		method toString() native
	}
```

PR #461, opened 2015-12-03
Dev #369 closures to wollok

javierfernandes:
Closures are now developed in wollok itself.


npasserini:
Zarpado!

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -512,6 +520,16 @@ package lang {
...
 		override method internalToSmartString(alreadyShown) = start.toString() + ".." + end.toString()
 	}
+	
+	/**
+	 * @author jfernandes
+	 * @since 1.3
+	 * @noInstantiate
+	 */
+	class Closure {
+		method apply(parameters...) native
```
javierfernandes: eventually we will have more methods here like "curry" and maybe even some almost reflectives like "getParameterNames()" ?

javierfernandes:
That was the last native object migrated to wollok.
We have now the path clear to perform a big semantic refactoring in the interpreter.
All evaluations should change from this:

```xtend
    def Object evaluate(EObject e)
```

To

```xtend
    def WollokObject evaluate(EObject e)
```

Also we can delete all the code for handling java ints, strings, etc.. in the runtime.
There should be only WollokObject.
All arguments to methods calls can also be considered as WollokObjects.

That should reduce the code.

npasserini:
Great!

PR #466, opened 2015-12-06
Dev #452 upgrade xtext 2.9

javierfernandes:


javierfernandes:
I had to disabled macos-32bits product building.
It failed in travis. I'm not sure why. I guess that there's an issue in the eclipse update sites.

PR #516, opened 2015-12-25
Feature unified block syntax

### LINKED ISSUE #415: Use Haskell/Prolog lists syntax
Most OO modern languages - JS, Python, Ruby, Swift, Groovy - use plain brackets to create literal collections, only XTend has the `#[]` syntax. The more symbols you need to represent a thing, the more ackward to the student looks. 

Also Haskell and Prolog have the same syntax for non head-tail lists. In order to keep compatibility with those languages, as in #409 in am proposing there to use that syntax.

flbulgarelli:
This PR introduces some important changes to syntax: 

* It unifies the code block syntax - only braces. This applies to block expressions, closures, etc
   * In particular, it changes the XTend syntax to `{ arg1,...argn => body }`, which resembles or is identical to Haskell's, Groovy, Scala, JS, etc
   * <del>It removes the optional arrow - now arrow is mandatory, to avoid ambiguities </del>
* It changes the list syntax to just `[1, 2, 3]`, to be compatible with the list syntax present in most modern languages 
* <del>It removes the Set literal syntax. But</del> it adds a `new Set(1, 2, 3)` constructor. 
* It removes the block as a top level expression. So it means you will not be able to do things like `var x = {2}`, only you will be able to use it in if/try expressions. 
* Due to ambiguites, now the scala-like method assignment syntax will only work for one liners. For example, instead of writing: 
* <del>`->` was removed from object. It is seldom used alongside the codebase, and it clashed with the arrow syntax for closures. If we really need tuples, perhaps we should add a better, more generic syntax.</del> 

```
method foo(x) = {
     y.z()
     4
}
```

You will have to write it this way:

```
method foo(x) {
     y.z()
     return 4
}
```

Which I think is actually a feature too, since it is very easy to ignore the equals sign between the parenthesis and brace.

You can still write this: 

```
method foo(x) = x
```

Fixes  #415

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -34,9 +34,9 @@ package lang {
 			printer.println(preffix +  this.className() + (if (message != null) (": " + message.toString()) else "")
 			
 			// TODO: eventually we will need a stringbuffer or something to avoid memory consumption
-			this.getStackTrace().forEach[e|
+			this.getStackTrace().forEach { e =>
 				printer.println("\tat " + e.contextDescription() + " [" + e.location() + "]")
-			]
+			}
 			
 			if (cause != null)
 				cause.printStackTraceWithPreffix("Caused by: ", printer)
@@ -110,19 +110,19 @@ package lang {
 		
 		method equals(other) = this == other
 		
+		method randomBetween(start, end) native
+		
 		method ->(other) {
 			return new Pair(this, other)
 		}
-		
-		method randomBetween(start, end) native
-		
+
 		method toString() {
 			// TODO: should be a set
 			// return this.toSmartString(#{})
-			return this.toSmartString(#[])
+			return this.toSmartString([])
 		}
 		method toSmartString(alreadyShown) {
-			if (alreadyShown.exists[e| e.identity() == this.identity()] ) { 
+			if (alreadyShown.exists { e => e.identity() == this.identity() } ) { 
 				return this.kindName() 
 			}
 			else {
@@ -132,9 +132,9 @@ package lang {
 		} 
 		method internalToSmartString(alreadyShown) {
 			return this.kindName() + "[" 
-				+ this.instanceVariables().map[v| 
+				+ this.instanceVariables().map { v => 
 					v.name() + "=" + v.valueToSmartString(alreadyShown)
-				].join(', ') 
+				}.join(', ') 
 			+ "]"
 		}
 		
@@ -145,7 +145,7 @@ package lang {
 					 	this.kindName()
 			message += " does not understand " + name
 			if (parameters.size() > 0)
-				message += "(" + (0..(parameters.size()-1)).map[i| "p" + i].join(',') + ")"
+				message += "(" + (0..(parameters.size()-1)).map { i => "p" + i }.join(',') + ")"
 			else
 				message += "()"
 			throw new MessageNotUnderstoodException(message)
@@ -178,21 +178,21 @@ package lang {
 		  * The criteria is given by a closure that receives a single element as input (one of the element)
 		  * The closure must return a comparable value (something that understands the >, >= messages).
 		  * Example:
-		  *       #["a", "ab", "abc", "d" ].max[e| e.length() ]    ->  returns "abc"		 
+		  *       ["a", "ab", "abc", "d" ].max { e => e.length() }    =>  returns "abc"		 
 		  */
-		method max(closure) = this.absolute(closure, [a,b | a > b])
+		method max(closure) = this.absolute(closure, { a, b => a > b })
 		
 		/**
 		  * Returns the element that is considered to be/have the minimum value.
 		  * The criteria is given by a closure that receives a single element as input (one of the element)
 		  * The closure must return a comparable value (something that understands the >, >= messages).
 		  * Example:
-		  *       #["ab", "abc", "hello", "wollok world"].max[e| e.length() ]    ->  returns "wollok world"		 
+		  *       ["ab", "abc", "hello", "wollok world"].max { e => e.length() }    =>  returns "wollok world"		 
 		  */
-		method min(closure) = this.absolute(closure, [a,b | a < b])
+		method min(closure) = this.absolute(closure, { a, b => a < b} )
 		
 		method absolute(closure, criteria) {
-			val result = this.fold(null, [acc, e|
+			val result = this.fold(null, { acc, e =>
 				val n = closure.apply(e) 
 				if (acc == null)
 					new Pair(e, n)
@@ -202,7 +202,7 @@ package lang {
 					else
 						acc
 				}
-			])
+			})
 			return if (result == null) null else result.getX()
 		}
 		 
@@ -211,7 +211,7 @@ package lang {
 		/**
 		  * Adds all elements from the given collection parameter to this collection
 		  */
-		method addAll(elements) { elements.forEach[e| this.add(e) ] }
+		method addAll(elements) { elements.forEach { e => this.add(e) } }
 		
 		/** Tells whether this collection has no elements */
 		method isEmpty() = this.size() == 0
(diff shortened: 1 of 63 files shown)


flbulgarelli:
Perhaps I can preserve the original set syntax. I will give a try

flbulgarelli:
@javierfernandes @npasserini Please comment

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -318,13 +314,16 @@ package lang {
 	 * @since 1.3
 	 */	
 	class Set inherits Collection {
-	
-		override method newInstance() = #{}
+		new(elements ...) {
+			this.addAll(elements)
+		}
+		
+		override method newInstance() = [].asSet()
```
javierfernandes: I think it is better to directly instantiate the Set. Otherwise we are creating a list instance just to throw-away. In general I don't mind memory-consumption, but in this case this is like a base class for wollok. we can avoid it easily

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/ClosureTestCase.xtend
```diff
@@ -32,7 +22,7 @@ class ClosureTestCase extends AbstractWollokInterpreterTestCase {
 		'''
 		program p {
 			var to = "world"
-			val helloWorld = ["hello " + to ]
+			val helloWorld = {->"hello " + to }
```
javierfernandes: This is the ugly side of the change :(

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/namedobjects/NamedObjectInheritanceTest.xtend
```diff
@@ -19,7 +19,7 @@ class NamedObjectInheritanceTest extends AbstractWollokInterpreterTestCase {
 			
 			program p {
 				assert.equals("abc", myObject.something())
-				assert.equals(myObject, (myObject->"42").getX())
+				assert.equals(true, (myObject == myObject))
```
javierfernandes: why was this changed ?
In this case the arrow is a message all objects understand that is currently used to create a pair/tuple.

We have a couple of special operators (in terms of the parser) that ends up being regular object messages.

org.uqbar.project.wollok/src/org/uqbar/project/wollok/WollokDsl.xtext
```diff
@@ -155,8 +155,7 @@ WOtherOperatorExpression returns WExpression:
 	rightOperand=WAdditiveExpression)*;
 
 OpOther:
-	  '->'
```
javierfernandes: mmm :(

org.uqbar.project.wollok.lib/src/wollok.wlk
flbulgarelli: You are right, I forgot to change it. Sorry!

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/ClosureTestCase.xtend
flbulgarelli: But the funny part is that the not arrow syntax was nearly not used in the whole wollok code base! I only fixed 2 or 3 occurrences I think. 

With the exception of Scala - which have partial application - and Ruby - well, when you don't use the arrow syntax -, I think having the no args arrow is pretty common.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/namedobjects/NamedObjectInheritanceTest.xtend
flbulgarelli: I added it on order to avoid ambiguities with  `->` in closures. 

Again, I nearly found no uses of `->` in the code base. Why is having literal pairs - and not is general concept of tuples - so important?

Anyway, another option is to use `=>` for creating tuples, or `=>` for closures. I prefer the first option, since it is far more common to use closures than tuples. What do you think?

flbulgarelli:
@npasserini @javierfernandes 

I did the necessary changes so that:
  * Pairs are re-introduced, by using => instead of -> for closures
  * Allow no arrow closures. 
 

But I am not being able to compile the type system projects anymore :/ I need your help. Could you please check checkout the branch and check the errors? 

Please don't merge anything into dev before this, I had to rebase this branch 4 or more times and since it is really big, it get reaaaaly tedious and error prone ....

flbulgarelli:
Ok, I could make it work. So it is ready to be merged.

javierfernandes:
Cool. I just built it locally fine.
Eclipse gets kind of crazy. 
When I move from one branch to another that has significant changes I do the following
* refresh all projects
* open the WollokDSL.xtext file and regenerate
* open the WollokDSL.xsemantics file change something and save (that regenerates xsemantics classes) 
* Project -> Clean (all)
* Maven -> Update Projects... (all)

With all that it should eventually get fully compiled.
Maven and travis should always break the tie.

I'll review now and merge

javierfernandes:
This renders the Language Reference outdated.
We need to update it.
And actually we might need to start splitting it into several pages depending on wollok version :S
Like Lang Ref 1.2, Lang Ref 1.3, etc.

flbulgarelli:
How can i update it?

javierfernandes:
try again now

npasserini:
I am very doubtful about the didactic implications of this change, I propose to revert and put a lot more thought on top of it.

This unifies things that are not necesarily the same... more over, they are explicitly handled differently in the grammar. 

Up to now, {} are used for blocks and for method bodies and [] for lambda expressions. This has nice consecuences for beginners, and resembles smalltalk syntax, which also differenciates those ideas.

In some languages these are similar things (most obviously Javascript) or have points in common (I think, Ruby, for example). But those features are either only useful for advanced programmers or even "hiding" the real nature of blocks from beginners.

Think Ruby's each: 

ary = [1,2,3,4,5]
ary.each do |i|
   puts i
end

Ruby as Wollok or Smalltalk, claim not having a loop construct, everything is about objects.
Still, the syntax tryes to hide it and a message sends looks very much like a loop construct.

Another example:

(1..10).each do |i|
 #  ... i goes from 1 to 10
end

So we have to make a decision. Are loop constructs good or bad? Is it better to have a message or not? If a message is better, why hide it? 
Of course I could imagine the answers to those questions, but... is it a good idea to say to a beginner to present concepts in that way? Talking about a theoretical purity which we want to hide a minute later? 

Will a student understand which are the advantages of those loops that are written as loops but shoud not be thougth as loops because they are in fact message sends? O will they just forget all the theoretical "nonsense" the teacher just said and use the good old loop construct that he already knows from a previous subject?


So, I think that this change has a didactical background which is making blocks less obvious and has didactical consequences which are  debatable. And well, once again I say we have to formalize a process to decide the shape of the language.

flbulgarelli:
> I am very doubtful about the didactic implications of this change, I propose to revert and put a lot more thought on top of it.

I agree that everything that must be discussed must be discussed now. So +1 to go on with the discussion. Nothing should be written in stone. 

But I disagree with strict reverting,  since we now have a clearer syntax for lists that does not collide with closures.  Let's continue the debate and understand the problems and solutions before taking further actions. 

> This unifies things that are not necesarily the same... more over, they are explicitly handled differently in the grammar.

Yes, that is true, but such comment also applies for class and method bodies. They are different things, handled different. 

However in both cases, as with closures, they share the notion of an aggregate of declarations or collaborations. 

> Up to now, {} are used for blocks and for method bodies and [] for lambda expressions. This has nice consecuences for beginners, and resembles smalltalk syntax, which also differenciates those ideas.

I would like to point out that blocks are not top level expressions anymore. That means, the following are closure, not block declarations now: 

```
method foo() = { 5 }
```

```
var x = { 5 }
```

Actually, a beginner would never notice that braces in a method declaration are blocks, she will find the braces as part of the rigid syntax of the language. That is, she can not explicitly create a block. Blocks have being keep as expressions only as part of the try-catch and if statements, just because of type system. 

So, to sum up, blocks are used only in the following scenarios:
  * method bodies
  * class/object bodies
  * if/try-catch statements. 

> In some languages these are similar things (most obviously Javascript) or have points in common (I think, Ruby, for example). But those features are either only useful for advanced programmers or even "hiding" the real nature of blocks from beginners.

Yes, that is true. Ruby/Scala/Groovy blocks where designed so that they resemble the syntax of a method/class body. The designers wanted DSLs to be seamless integrated into the language. And to encourage programmers to "extend" the language

*Side note: Java/C# closures are harder to mistake with blocks. But that is just because closure arrow is mandatory*

That decision also encourages the point of view that actually having a harcoded syntax and semantics for some aspects of the language is not really important - conceptually they could have being implemented with closures.

Actually, we are implementing if/try as native statements, but we now they could have being just messages that take closures. 

>Think Ruby's each:
>
>ary = [1,2,3,4,5]
>ary.each do |i|
>puts i
>end
>
>Ruby as Wollok or Smalltalk, claim not having a loop construct, everything is about objects.
>Still, the syntax tryes to hide it and a message sends looks very much like a loop construct.
>
>Another example:
>
>(1..10).each do |i|
># ... i goes from 1 to 10
>end
>

True. 

>So we have to make a decision. Are loop constructs good or bad? Is it better to have a message or >not? If a message is better, why hide it? 

I am not getting your point here :disappointed: 

> Of course I could imagine the answers to those questions, but... is it a good idea to say to a beginner to present concepts in that way? Talking about a theoretical purity which we want to hide a minute later?
> Will a student understand which are the advantages of those loops that are written as loops but shoud not be thougth as loops because they are in fact message sends? O will they just forget all the theoretical "nonsense" the teacher just said and use the good old loop construct that he already knows from a previous subject?

Again, I don't understand this statement. 

If you are saying that some students may not recognize that it is collaboration instead of a harcoded statement, I think the same applies to Smalltalk. Let's see and IfTrue: 

```smalltalk
(5 > counter) ifTrue: [
   self  notifyTimeout: client
]
```

I have noticed many students don't early realize that it is a collaboration instead of a harcoded syntax. And this is not because of a confusing syntax, because there is no such thing as *blocks* - in wollok terms - in Smalltalk, nor method or class delimiters. 

But I don't think this is a big problem. They eventually get it, but even if they didn't, PDEP is not an advanced course where we teach to model with closures. 

> So, I think that this change has a didactical background which is making blocks less obvious and has didactical consequences which are debatable.

Ok, let's debate it. Is this seamless good or bad? How good and how bad?

I have no strong position here right now. I think that the metaphor of "extending the language" / "blocks..closures.. don't care" is good. Moreover, if you remember that it only affects the two native constructs that may have being modeled with closures (if's and try's).

But, perhaps that could confuse a student, as you pointed out. I really don't know. We should test it. 

Perhaps instead of using a different syntax for closures, we should use a different syntax for bodies. Don't using braces for those statements is always a not too expensive option. For example, a try-catch nearly don't requires braces: 

```
try 
   ....
catch (Exception e)
   ...
end
```

*Side note:  could we buy http://dl.acm.org/citation.cfm?id=2534973? Those guys actually did research on syntax, and some of their conclusions apply to wollok. They results are [here](http://quorumlanguage.com/documents/syntax/classes.php)

>And well, once again I say we have to formalize a process to decide the shape of the language.

@npasserini As in a previous comment, I agree, but I also ask you: how?

flbulgarelli:
Just another alternative for keeping a clear separation between closures and bodies AND still keeping the clear separation between lists and closures is something in the middle of Ruby and JS: appending a token.

For example:

```
val double = lambda { x => x * 2 }
[3, 4, 5].map(lambda { x => x + 1 })
```

```
val double = closure { x => x * 2 }
[3, 4, 5].map(closure { x => x + 1 })
```

```
val double = closure(x) { x * 2 }
[3, 4, 5].map(closure(x) { x + 1 }) 
3.times(closure { 
   ...
})
```




This syntax is quite verbose, but I am sure it avoids ambiguities and avoids introducing new symbols. And is not so drastic as changing braces for `end`s.

I am not in love with it, but it addresses @npasserini 's doubts.. It could work. 

PD: @matifreyre suggested removing any kind of optional parenthesis. With this closure syntax, removing such option has more sense.

javierfernandes:
So the initial motivation was to replace this
```xtend
#[ 1, 2, 3 ]
```

with this
```xtend
[ 1, 2, 3]
```

And now we will end up writting
```
closure { blah }
```
instead of 
```
[ blah ]
```

I don't like changing the body's braces to "begin/end", or just "end".
I knew we will end up at this point... ruby syntax. I feel kind of a boiled frog :)

npasserini:
My point is that we are making two different things to look the same just
getting rid of a # in the list literal.
We are adding confusion and we are loosing didactical clarity of our
language.

npasserini:
On Sat, Jan 2, 2016 at 3:53 AM, Franco Leonardo Bulgarelli <
notifications@github.com> wrote:

> But I disagree with strict reverting, since we now have a clearer syntax
> for lists that does not collide with closures. Let's continue the debate
> and understand the problems and solutions before taking further actions.
>
>
And my point is that the debate you now propose should have had place
before changing the language and not after introducing a disruptive change
with didactical implications.

flbulgarelli:
:ok_hand:

flbulgarelli:
PS: The issue discussing this change - which I opened 45 days ago and was nearly ignored - is #414. 

However, I will open a PR for reverting those changes as request - I am currently working on it -, and then open a new issue for general syntax change, with hope that discussing in more general terms instead of particular features motives people to debate, gain consensus and establish a more visible language discussion protocol  

I will also republish my design criteria proposal I already sent to the dev list, with faith in finally debating about it instead of reducing the discussion to counting characters and personal tastes. This is not the kind of syntax and semantic discussion I would like to hold.

PR #520, opened 2015-12-25
Issue #409 preludes collection messages

### LINKED ISSUE #409: Use collection messages consistent with Haskell's Prelude
The following collection messages are inconsistent with Haskell's Prelude

* `detect` -> `find`
* `exists` -> `any`
* `forAll` -> `all`

flbulgarelli:
Merge after #516 

Fixes #409


flbulgarelli:
Merging this since it seems to be partial agreement about those names. Perhaps there is not absolute agreement about `find`, but we anyway we need to continue working on its naming since #522

PR #525, opened 2015-12-26
Feature #387 dictionary syntax

### LINKED ISSUE #387: Add Map/Dictionary objects
With a literal expression

flbulgarelli:
Merge after #519

Fixes #387

It is actually not finished yet - there are some tests failing and not all important messages have being implemented - but I would like some feedback before proceeding.


flbulgarelli:
@npasserini @javierfernandes

org.uqbar.project.wollok.launch/src/org/uqbar/project/wollok/debugger/server/rmi/XWollokObjectDebugValue.xtend
```diff
@@ -41,6 +41,7 @@ class XWollokObjectDebugValue extends XDebugValue {
 	}
 	
 	def isList() { typeName == LIST }
+	def isDictionary() { typeName == DICTIONARY }
```
javierfernandes: Eventually we will add a custom view in the objects diagram to show dictionaries, maybe like a 2 columns table, or something (needs to be more clever, to decide if it will put the value and key object within the table or just an arrow to an object outside the table.
#526 new issue

org.uqbar.project.wollok.lib/src/wollok/lang/WDictionary.xtend
```diff
@@ -0,0 +1,72 @@
...
+
+	def void clear() { wrapped.clear }
+
+	@NativeMessage("equals")
+	def wollokEquals(WollokObject other) {
+		other.hasNativeType(this.class.name) && (other.getNativeObject(this.class).wrapped == this.wrapped)
+	}
+
+	@NativeMessage("==")
+	def wollokEqualsEquals(WollokObject other) { wollokEquals(other) }
```
javierfernandes: This can be written in wollok itself. It just delegates.
I believe it's already implemented like that in wollok.lang.Object.
So only one should be overriden natively.

Actually I don't think we need to re implement the equals natively, if it is already implemented in Object based on identity

org.uqbar.project.wollok.lib/src/wollok/lang/WDictionary.xtend
```diff
@@ -0,0 +1,72 @@
...
+	
+	def Object values() {
+		wrapped.values().toList
+	}
+	
+	def Object keys() {
+		wrapped.keySet()
+	}
+	
+	def Object filter(WollokObject proc) {
```
javierfernandes: If we expose an "entrySet()" message, coded natively, then "filter" can be implemented in wollok based on that method.

org.uqbar.project.wollok.lib/src/wollok/lang/WDictionary.xtend
```diff
@@ -0,0 +1,72 @@
...
+		wrapped.entrySet.forEach[ 
+			if (closure.apply(it.value).wollokToJava(Boolean) as Boolean) {
+				dict.put(it.key, it.value)
+			}
+		]
+		
+		dict
+	}
+	
+	def Object map(WollokObject proc) {
```
javierfernandes: Same

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/DictionaryTestCase.xtend
```diff
@@ -0,0 +1,231 @@
...
+import org.junit.Test
+import org.uqbar.project.wollok.interpreter.core.WollokProgramExceptionWrapper
+
+/**
+ * @author jfernandes
+ */
+class DictionaryTestCase extends AbstractWollokInterpreterTestCase {
+	
+	def instantiateObjectsDictionary() {
+		"val dictionary = {foo:1, bar:'hello', baz: {:}}"
```
javierfernandes: ```{ : }``` seems weird.
What would be that for ?
Can we just forbid that at grammar level ?

javierfernandes:
Sounds good. We needed dictionaries.

Remember to update the Reference Guide (https://github.com/uqbar-project/wollok/wiki/LanguageReference) and https://github.com/uqbar-project/wollok/wiki/Wollok-v1.4-Diderot once tests pass and we have merged this into dev

org.uqbar.project.wollok.lib/src/wollok/lang/WDictionary.xtend
flbulgarelli: I take it from the WCollection class. Should I remove from that class, too?

org.uqbar.project.wollok.lib/src/wollok/lang/WDictionary.xtend
flbulgarelli: But that would retrieve a set of entries, not a dictionary with filtered elements. 

With this proposal you can:
   * map/filter the dictionary
   * map/filter the keys
   * map/filter the values

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/DictionaryTestCase.xtend
flbulgarelli: I had to add it in order to avoid ambiguities with the expression block. Otherwise the following would be ambiguous:
```
if (true) 
  {}
```

Ruby doesn't have that issue since they don't use braces. And Groovy adopted the same solution (an empty dictionary in Groovy is `[:]`)

But I am open to alternatives here, Ink it can be done better. What do you suggest?

PR #550, opened 2015-12-30
the main thread join the mainLoopThread until it ends

### LINKED ISSUE #543: System.exit(0) breaks WollokGame
Since it ends the whole process killing libgdx background thread.

Wollok game start() method should block the thread. And maybe continue it once libgdx notifies that the window has been closed.

Juancete:
Create a new LwjglApplication class. The main thread join to LwjglApplication thread and waits until it ends.

Close #543


PalumboN:
:cool:

PR #551, opened 2015-12-30
Replace config.json with wollok code

PalumboN:
Close #485

PR #552, opened 2015-12-30
Modeling keyboard keys

PalumboN:
- Now each key is an object, located in game package, that know its key code.

- I added only the common keys and not all framework options.

- I grouped  similar keys in one object. For example: SHIFT, ALT & CONTROL work for both left and right keys.

Merge after https://github.com/uqbar-project/wollok/pull/551


Juancete:
:cocktail: :+1: 

We need to separate wollok game code from wollok.wlk in more files

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -689,7 +668,227 @@ package lib {
...
+	
+	object X inherits Key {		
+		method getKeyCodes() = [52]
+	}
+	
+	object Y inherits Key {		
+		method getKeyCodes() = [53]
+	}
+	
+	object Z inherits Key {
```
flbulgarelli: I think some is asking for enum's here

javierfernandes: This can be less verbose right now by calling a superclass constuctor like
this

```xtend
object W inhertis Key(51)
object X inhertis Key(52)
object Y inhertis Key(53)
```

Preethe close to an enum

PalumboN: Oohh nice!

PR #557, opened 2015-12-30
Fix #511 allow to override object methods

### LINKED ISSUE #511: Classes and objects can't override methods to Object class by default
For example
```Xtend
class Foo {
	override method == (other) {
		return true
	}
}
```
doesn't compile because _Method does not override anything_.

However if I write the `inherits`
```Xtend
class Foo inherits Object {
	override method == (other) {
		return true
	}
}
```
it's ok.

npasserini:


javierfernandes:
I will need to merge this with mixins

PR #560, opened 2015-12-31
Introducing constructor token

### LINKED ISSUE #527: Change "new" keyword to "constructor" for constructors declarations
Currently, constructors are declarared using the new keyword.

This is confusing, since we are using the same token for very different use cases. And it does not match the name we use for it: constructor. 

Also, since clases are not objects, the intuition of being a method and thus shares the selector mesaage is broken.

JS does it better: it has a constructor keyword. 

My proposal is to rename the **new** token to **constructor**. 

@npasserini this is what we discussed

flbulgarelli:
Fixes #527


npasserini:
:+1:

Issue #541, opened 2015-12-28
Design an icon for Mixins in label providers


javierfernandes:
Autocomplete, outline, etc.

<img width="265" alt="screen shot 2015-12-28 at 01 43 42" src="https://cloud.githubusercontent.com/assets/4428120/12014119/88299db8-ad04-11e5-8eab-eb42d74809b9.png">

It is currently being shown with the default icon

javierfernandes:
New icon. This is how it looks


![uploads/88ea33dc-6993-4416-94b2-0f2dfdd5e49d/Screen Shot 2015-12-30 at 14.28.26.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F88ea33dc-6993-4416-94b2-0f2dfdd5e49d%2FScreen+Shot+2015-12-30+at+14.28.26.png)

PalumboN:
Like!

Issue #540, opened 2015-12-28
Show mixins in static diagram


javierfernandes:

javierfernandes:
The diagram layout sucks, and the code is terrible, since it is really hacked-up and has duplication.
All the diagrams need a really good refactoring, but well, mixins are now shown

<img width="401" alt="screen shot 2016-01-03 at 20 57 48" src="https://cloud.githubusercontent.com/assets/4428120/12105557/a910ff0e-b334-11e5-9658-55570ab8318e.png">

npasserini:
Great

Issue #538, opened 2015-12-27
Statically check consistent hierarchy when instantiating


javierfernandes:
Combining abstract mixins means that you can end up with a hierarchy which is inconsistent, if no element defines an the expected abstract method

Example

```scala
mixin M1 {
    method m() {
             this.doIt()      // expected abstract method
    }
}

class C mixed with M1   // this class should be considered abstract !
```

Then trying to instantiate it

```scala
val c = new C()
```
Should not compile telling the class is abstract since it doesn't implement the required abstract method "doIt()"

Issue #537, opened 2015-12-27
Allow to combine mixins at instantiation time


javierfernandes:
To avoid generating subclasses just to combine mixins it would be good to allow instantiation-time mixins combination, in the same way Scala allows.

```scala
mixin M1
mixin M2

class A 
```

Could be used like this

```scala
val a = new A() mixed with M2, M1
```

This produces a hierarchy like this

```bash
? -> M2 -> M1 -> A
```
(depending on the syntax #536)

javierfernandes:
This is implemented in 1.6

PR #561, opened 2015-12-31
Dev mixins

javierfernandes:
Solves #322, #541, #538, #535 , #154

Future discussions: #536, #537, #540


org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/mixins/MixinsTestCase.xtend
```diff
@@ -0,0 +1,369 @@
...
+		program t {
+				val b = new WalkingBird()
+				b.walk(10)
+				assert.equals(10, b.walkedDistance())
+		}
+		'''.interpretPropagatingErrors
+	}
+	
+	// is it ok to have 'protected' variables ?
+	// this is inconsistent with the way wollok handles variables in classes in a hierarchy
```
npasserini: Why? Aren't all variables 'protected'?

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/mixins/MixinsTestCase.xtend
```diff
@@ -0,0 +1,369 @@
...
+			method doFoo(chain) { super(chain + "> M2") }
+		}
+		
+		class C1 {
+			var foo = ""
+			method doFoo(chain) { foo = chain + " > C1" }
+			method foo() = foo
+		}
+		
+		class C2 inherits C1 mixed with M1, M2 {
```
npasserini: Here you use 'mixed with M1, M2' while before you have used 'mixed with M1 and M2', are both right?

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/xpect/validations/mixins/MixinInstantiation.wlk.xt
```diff
@@ -0,0 +1,91 @@
...
+
+class H inherits DoItClass mixed with M2 {}
+
+class I mixed with M2 {
+	method doIt() {}
+}
+
+object p {
+	method run() {
+		// XPECT errors --> "Must implement inherited abstract methods: M1.doIt()" at "C"
```
npasserini: I think we should add 'C can not be instantiated. ' to this error message.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/xpect/validations/mixins/MixinInstantiation.wlk.xt
```diff
@@ -0,0 +1,91 @@
...
+		val c = new C()
+		console.println(c)
+		
+		// XPECT errors --> "Must implement inherited abstract methods: M1.doIt()" at "D"
```
npasserini: Why not replace 'M1.doIt()' with 'doIt() (required by M1)

I like the term "required" as more natural dan "abstract".

org.uqbar.project.wollok/src/org/uqbar/project/wollok/WollokDsl.xtext
```diff
@@ -41,14 +41,17 @@ WPackage:
...
+WNamed: WMethodDeclaration | WReferenciable | WPackage | WClass | WMixin;
+WMethodContainer: WClass | {WObjectLiteral} | WNamedObject | WMixin;
 WReferenciable: WParameter | {WVariable} | WNamedObject ;
 
 WObjectLiteral returns WExpression:
 	{WObjectLiteral}
-	=>'object' '{'
+	=>'object'
+	('inherits' parent=[WClass|QualifiedName] ('(' parentParameters+=WExpression (',' parentParameters+=WExpression)* ')')? )?
+	('mixed with' mixins+=[WMixin|QualifiedName] (('and'|',') mixins+=[WMixin|QualifiedName])*)?
```
npasserini: What about replacing 'mixed with' with 'uses'

org.uqbar.project.wollok/src/org/uqbar/project/wollok/WollokDsl.xtext
```diff
@@ -324,13 +330,25 @@ WCanExtendClass:
...
 		(members+=WVariableDeclaration ';'?)*
 		(constructors+=WConstructor)*
 		(members+=WMethodDeclaration ';'?)*
 	'}'
 ;
 
+// MIXINS
+
+WMixin:
+	'mixin' name=ID '{'
```
npasserini: I do not love the word 'mixin', I think it means nothing for the student.
What about using a more significant name?... 'behavior' could be a name.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/mixins/MixinsTestCase.xtend
javierfernandes: Not currently.
Instance variables in classes are private.
But a subclass cannot declare a new one with the same name as one in a superclass.
Although is just a static check.

Mixins instance variables on the other hand was implemented as protected.

I woudln't be a problem to change class variables to protected.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/mixins/MixinsTestCase.xtend
javierfernandes: Yes, you can use comma or "and"
I found "and" more human readable and comma better if you want to write it faster and less verbose.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/xpect/validations/mixins/MixinInstantiation.wlk.xt
javierfernandes: Yes, Actually it was just saying that. It now shows the methods that it needs to implement, but I lost the first part of the message :P

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/xpect/validations/mixins/MixinInstantiation.wlk.xt
javierfernandes: Yes. I'm just concern about cases where you have many errors, the message will be quite long. But, maybe that doesn't matter.

org.uqbar.project.wollok/src/org/uqbar/project/wollok/WollokDsl.xtext
javierfernandes: Well, actually franco didn't like "mixed with". He said that one commonly uses the "mix" verb to refer to the mixing as subject like "Mixin1 mixed in ..." and not the other way.
I'm not sure about that.
He proposed the "includes" which is the same as ruby uses.
I think that both "uses" and "includes" are two ambiguous.
If we follow the same reasoning as we used for changing "extends" to "inherits" we should find a term that matches the concrete case for mixins. That's why I used the "mixed in" phrase, because explicitely tells we talking about mixins.

What if we find another meaning to "uses" or "includes" in the future ?
Also we use to talk about an object "using" another object as collaboration. So it could be confusing.
"Includes" is also used in template engines and other stuff.

org.uqbar.project.wollok/src/org/uqbar/project/wollok/interpreter/WollokInterpreterEvaluator.xtend
```diff
@@ -284,16 +285,20 @@ class WollokInterpreterEvaluator implements XInterpreterEvaluator {
 	}
 	
 	def createNamedObject(WNamedObject namedObject, String qualifiedName) {
-		namedObject.hookObjectInHierarhcy(classFinder)
+		namedObject.hookObjectInHierarchy(classFinder)
```
npasserini: Is this necessary? 
Didn't we remove this?

org.uqbar.project.wollok/src/org/uqbar/project/wollok/interpreter/WollokInterpreterEvaluator.xtend
```diff
@@ -313,6 +320,15 @@ class WollokInterpreterEvaluator implements XInterpreterEvaluator {
...
 	}
+	def static hookObjectInHierarchy(WObjectLiteral namedObject, WollokClassFinder finder) {
+		if (namedObject.parent != null)
+			namedObject.parent.hookToObject(finder)
+		else {
+			val object = finder.getObjectClass(namedObject)
+			namedObject.parent = object
+			namedObject.eSet(WollokDslPackage.Literals.WOBJECT_LITERAL__PARENT, object)
+		}
+	}
```
npasserini: All this code should not be necessary any more.

org.uqbar.project.wollok/src/org/uqbar/project/wollok/WollokDsl.xtext
javierfernandes: Ok, this one is heavy.
I agree that mixin is a general keyword that doesn't say anything.
On the other hand it is the most used term in programming literature.

I cannot think of a better term. I'm afraid that **behaviour** will collide with the speech when we say that an object has behaviour besides data.

As a **mixin** can model anything, it will be hard to find a good name.
Lets think. It doesn't depend on its content.
It could have more than one method, so it could add more than one behaviour to an object.
Maybe it is like a **feature** or **characteristic**, 
Sometimes it is used to model a **role**, although it seems to be a name which doesn't apply to all cases.
The main distinction from a class is that it cannot be instantiated, so it should model **part** of the behaviour of an object.
Maybe it is a **partial** ?

org.uqbar.project.wollok/src/org/uqbar/project/wollok/model/WMethodContainerExtensions.xtend
```diff
@@ -130,35 +166,66 @@ class WMethodContainerExtensions extends WollokModelExtensions {
...
+	def static List<WMethodContainer> linearizateHierarhcy(WMethodContainer it) {
+		var chain = newLinkedList
+		chain.add(it)
+		if (mixins != null) {
+			val reversed = mixins.clone.reverse
+			chain.addAll(reversed)
+		}
+		if (parent != null)
+			chain.addAll(parent.linearizateHierarhcy)
+		chain
```
npasserini: What happens if a mixin appears more than once in the chain?

org.uqbar.project.wollok/src/org/uqbar/project/wollok/interpreter/WollokInterpreterEvaluator.xtend
javierfernandes: No, and yes.
Just that dev-1.4 and mixins branches were not updated yet.
Once I merge mixins into dev-1.4 I will merge dev -> dev1.4 and will be removed

org.uqbar.project.wollok/src/org/uqbar/project/wollok/model/WollokModelExtensions.xtend
```diff
@@ -177,11 +180,13 @@ class WollokModelExtensions {
...
 	
-	def static isValidCallToWKObject(WMemberFeatureCall it, WollokClassFinder finder) { resolveWKO(finder).isValidCall(it, finder) }
+	def static isValidCallToWKObject(WMemberFeatureCall it, WollokClassFinder finder) { 
+		resolveWKO(finder).isValidCall(it, finder)
+	}
 	
 	def static resolveWKO(WMemberFeatureCall it, WollokClassFinder finder) { 
 		val obj = (memberCallTarget as WVariableReference).ref as WNamedObject
-		obj.hookObjectInHierarhcy(finder)
+		obj.hookObjectInHierarchy(finder)
```
npasserini: Again should not be necessary.

org.uqbar.project.xinterpreter/src/org/uqbar/project/wollok/interpreter/api/XInterpreterEvaluator.xtend
```diff
@@ -18,10 +17,11 @@ import org.uqbar.project.wollok.interpreter.core.WollokObject
  * 
  * @author jfernandes
  */
-interface XInterpreterEvaluator extends Serializable {
+interface XInterpreterEvaluator<O> extends Serializable {
```
npasserini: Why is this necessary?

javierfernandes: I have abstracted part of the interpreter into a general project called xinterpreter. Therefore the resulto of evaluating depends on the runtime model of the language. In our case WollokObject.

org.uqbar.project.wollok/src/org/uqbar/project/wollok/WollokDsl.xtext
npasserini: "chunk"!

PR #565, opened 2016-01-01
Adding findOrElse and findOrDefault

### LINKED ISSUE #522: find should fail when value is not present
Returning null is not inherently bad, but leads student into null checking, which is a practice we should avoid. 

I propose to simply fail on this scenario, or add a variant with an extra continuation for the fail scenario, or  return a maybe functor.

### LINKED ISSUE #567: Add assert.throwsException
Add a method for checking that a no arg closure, when evaluated, actually throws an exception <del>of an specific type.</del>

flbulgarelli:
Fixes #567 
Fixes #522

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -596,6 +596,7 @@ package lib {
 		method notThat(value) native
 		method equals(expected, actual) native
 		method notEquals(expected, actual) native
+		method throwsException(block) native
 		method fail(message) native
 	}
 	
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -251,7 +251,7 @@ package lang {
 		 * If more than one element satisfies the condition then it depends on the specific collection class which element
 		 * will be returned
 		 * @returns the element that complies the condition
-		 * @throws
+		 * @throws ElementNotFoundException if no element matched the given predicate
 		 * Example:
 		 *      users.find { user => user.name() == "Cosme Fulanito" }
 		 */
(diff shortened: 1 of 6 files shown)


org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -246,15 +251,35 @@ package lang {
...
+		/**
+		 * Returns the element of this collection that satisfy a given condition, 
+		 * or the the result of evaluating the given continuation 
+		 * If more than one element satisfies the condition then it depends on the specific collection class which element
+		 * will be returned
+		 * @returns the element that complies the condition or the result of evaluating the continuation
+		 * Example:
+		 *      users.findOrElse({ user => user.name() == "Cosme Fulanito" }, { homer })
+		 */
+		method findOrElse(predicate, continuation) native
```
javierfernandes: Couldn't this be implemented in wollok ?

flbulgarelli: I fear not. See my comment in 0759764aa24e4dd4ccf39a8c3b944a342b40ad48

> Since we have no non-local-returns, and the fold solution is not null safe and requires the whole collection to be iterated

flbulgarelli: I actually tried the following manners:

* using a wollok fold. That works, but it requires the whole collection to be iterated - no short circuit -, and that requires the continuation to never return null. It makes the contract more complex
* using a wollok forEach. That could work, but you can not return from the method within the closure, since closure return only escapes from the closure. It is a local return

Have I misunderstood something?

flbulgarelli: PS: I am wondering that not having non-local returns AND no for statement AND no streams  means means that there are many algorithms that can not be implemented directly in the language, right?

npasserini: Is there any contraindication for the non local returns?
(I mean, besides the implementation complexity.)

flbulgarelli: I am not against them :smile: 

What is more, I am starting to think they are necessary.

PR #573, opened 2016-01-04
Reverting feature-unified-block-syntax

flbulgarelli:
Reverts #516


npasserini:
Which is the difference between this and #572?

flbulgarelli:
Finally we decided to not rollback this issue

PR #595, opened 2016-03-04
Swap height and width iterations.

### LINKED ISSUE #594: [wollokgame] The tile image won't fill the screen when the window is rectangular
If you define a rectangular Gameboard, the tile image is not replicated as it should. For example:

 ```
program example {
	wgame.setHeight(4)
	wgame.setWidth(11)
        ....
 ```

Seems to be swapped height and width iterations in some place.

Juancete:
Close #594 

Just swap the height and width iterations when the tile image is created.


PalumboN:
And it works for all Gameboard configurations? For example `4x11` & `11x4`?

PalumboN:
:cocktail:

Juancete:
Yeap... works with 4x11 & 11x4 also :D

PR #596, opened 2016-03-05
Fix #507 Double-clicking on a test raises a Java null pointer exception

### LINKED ISSUE #507: Double-clicking on a test raises a Java null pointer exception
When double-clicking **on a test that resulted in error**, instead of the expected navigation to the corresponding test, a Java exception is raised.

image

npasserini:
Temporarily disable linking to the URI of the error source, since it is not available.
Until that is fixed I provide a link to test header.


javierfernandes:
Is it always null ? or just for some kind of errors ?

PR #607, opened 2016-03-07
Fixes #188 Formater should keep sequences of commands in separate lines.

### LINKED ISSUE #188: Formater should keep sequences of commands in separate lines
For example in the following program, current formatter brings both println's in the same line.

object foo {
    method bar() {
        console.println("")
        console.println("")
    }
}

npasserini:


org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/xpect/formatter/WollokFormatterTestCase.xtend
```diff
@@ -55,6 +55,7 @@ class WollokFormatterTestCase extends AbstractXtextTests {
     }
     
    	def assertFormatting(String program, String expected) {
+   		println(formatCode(program))
```
javierfernandes: to remove on further commits :P

javierfernandes:
Eventually we need to add more tests for the formatter. Since it is really sensible to changes.
I've read that xtext new version has a new API for formatting that's why I haven't pay much attention to this issues (also because they are like low priority compared to some others).
But well, each change to the grammar also impacts the formatting, so we should create new tests.
This is one of many motivators to create a webapp so that people can contribute with testcases without having the full wollok dev environment :S

npasserini:
The xtext new version is the one we are using?
Or should we migrate to get that advantage?

Issue #510, opened 2015-12-22
Should be a way to have an ordered list


PalumboN:
Sometimes the exercises in the guides require order a list by a condition. It would be great if Wollok could support it.

matifreyre:
+1

javierfernandes:
Ok there could be two different scenarios/requirements
1. **to sort a list*
1. to have a **sorted collection**

npasserini:
Thinking on most common teaching examples, I think that sorting a list
covers pretty much all the exercises we usually use, and is more simple to
teach than the other possibility.

matifreyre:
You mean sorting the list with side-effect or create a new list with its elements sorted? 
I prefer the latter.

npasserini:
I am not sure about what is best.
Maybe you could reorder the same list and force a copy only if you need it.

matifreyre:
It's for didactic reasons, I prefer creating a brand new list.
To begin with, it's the same you do in Smalltalk with ```asSortedCollection:```, but it's also what you would do in Haskell and Prolog. Of course, you don't have the alternative of reordering the list there, but I feel like it's more consistent to do the same in Wollok. 
We can also have 2 different messages like ```sortBy(aClosure)``` and ```asSortedBy(aClosure)```.

flbulgarelli:
For didactic purposes, a pure `sort`/`sortBy` is enough. Having an in-place sorting alternative is plus.

I didn't get the different between `sort` and `asSorted`, which was the intention?

matifreyre:
Reorder the same collection (side - effect) or create a new one (no side
effect) respectively.

matifreyre:
Any thought's on this? My proposal was to have 2 separate messages:
```sort```: changes the receiving collection.
```asSorted```: creates and returns a new collection.
In both cases we could add a *```By```* at the end.

flbulgarelli:
I like to have the four methods you mention. Just I am not conviced of the names, I think it is very related to #517 , because it depends on the decision about having a convention or not to mark imperative-style, effectful methods, 

By the way, I think we can get with @matifreyre naming proposal and then rename if needed.

asanzo:
I think the 4 method approach is elegant and suits every need. 
sort / asSorted / sortBy / asSortedBy 
You could teach using whichever you like.

matifreyre:
I'm working on this. I'm having some problems with conversions between Java and Wollok objects, I just need some more time to find out where that conversion is being made.

[...]

PR #610, opened 2016-03-08
Adding sorting methods to collections

matifreyre:
Feature #510: Should be a way to have an ordered list
Adding sortBy to List (side-effect version) and sortedBy to Collection (no side-effect version: returns a new List).
Please let me know everything you see that can be improved.

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -333,6 +333,12 @@ package lang {
 		method toStringSufix()
 		method asList()
 		method asSet()
+		method copy() {
+			var copy = this.newInstance()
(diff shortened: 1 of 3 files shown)


matifreyre:
Those tests run smoothly on eclipse. I'm not sure what's making them fail here.

javierfernandes:
The build in travis was failing due to some sort of transient error related to network connections to maven repo's or something.
It now ran but I got some errors in ListTestCase. The new tests include "val" keyword which should now be "const".
I'm changing that.

javierfernandes:
ListTest seems to fail randomly (the push build succeded by the PR build failed)

javierfernandes:
Ran fine locally. Let's cross fingers

Issue #603, opened 2016-03-07
WKOs that inherit from a class can't use inherited attributes in its code


mmatos:
Unlike subclasses, methods in a WKO that inherit from a class that defines an attribute cannot use that attribute directly. The code doesn't compile and the error "Couldn't resolve reference to Referenciable <attribute name>" appears.

javierfernandes:
Ok. So the issue here is not in the runtime part of wollok (the interpreter), but in the static part.
This check is not coded by us, but an out-of-the-box check that xtext provides for references (the linker).
This linker bases the reference resolution and the scoping of your language.
The scope can be customized, and it is in Wollok.

I don't remember the exact name of the class, there are a couple of "Wollok*Scope*" classes.
Some of them are for resolving imports and more hardcore stuff. But there's one that is coded in a declarative way following xtext convention that should have definitions of scopes for, for example a Method (WMethodDeclaration), and Class (WClass), and so on.

I guess that you should customize there the scoping for WNamedObject (WKO's) probably following what is already there for WClass'es

Let me know if this sounds reasonable

mmatos:
Thanks a lot for the help!
It would seem that this is the cause, in WollokDslScopeProvider
	def dispatch IScope scope(WNamedObject namedObject) {
		namedObject.declaredVariables.asScope
	}
The parent is not taken into consideration, unlike the definition for WClass

javierfernandes:
Solved right ?

PR #618, opened 2016-03-10
#603 Fix for variable scope of NamedObject

mmatos:
For the moment, I only changed scope(WNamedObject namedObject) to consider the parent's scope as well. I didn't include mixins because I wasn't sure about what was expected on that aspect (not sure if NamedObjects will be expected to be mixed in in the future).

All preexistent test cases pass and the additioned objectInheritFromCustomClassWithInstanceVariable test case in NamedObjectInheritanceTest is also green after the fix.

In an unrelated topic, after I made the change I noticed that the following files were also updated and git prompted me to include them in the commit, but I think they just need to be ignored and not commited (they are listed in .gitignore files under org.uqbar.project.wollok.typeSystem.xsemantics\xsemantics-gen\org\uqbar\project\wollok\semantics):
/.WollokDslTypeSystem.java._trace
/.WollokDslTypeSystemValidator.java._trace

Please advice in case I need to do something about that.

### DIFF
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/mixins/MixinsTestCase.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/mixins/MixinsTestCase.xtend
@@ -303,6 +303,31 @@ class MixinsTestCase extends AbstractWollokInterpreterTestCase {
 	}
 	
 	@Test
+	def void mixinVariablesAreInScopeOnAWKO() {
+		'''
+		mixin Flies {
+			var times = 0
+			method fly() {
+				times = 1
+			}
+			method times() = times
+		}
+		
+		object pepita mixed with Flies {
+			method rest() {
+				times = 0
+			}
+		}
+		
+		program t {
+			pepita.fly()
+			assert.equals(1, pepita.times())
+			pepita.rest()
+			assert.equals(0, pepita.times())
+		} 
+		'''.interpretPropagatingErrors
+	}
+	@Test
 	def void mixinOnAWKOOverridingAMethod() {
 		'''
 		mixin Flies {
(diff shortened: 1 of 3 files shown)


javierfernandes:
Yes. trace files are generated by xtend compiler.
We don't need to version them. Maybe we did it already, then we should clean up.

Regarding your change. It seems fine.
Just that, WKO already support mixins, and they are currently being taking into account for the scope of WClass (classes), so WKO are more close to be consistent with classes with your change, but not 100%.

Do you mind also adding the mixins scope ?
At least we will have something 100% consistent.
Then we can discuss to change the scoping as we want.

Agree ?

mmatos:
Yes, I agree, just wanted to confirm first

mmatos:
I added the change to consider also mixins and a test case to validate it works as expected, though there is duplicated code.

Looking at WNamedObject and WClass I see that they both define getParent and getMixins independently, so maybe those declarations could be in WCanExtendClass so a single definition for scope can be done for WCanExtendClass?

javierfernandes:
Cool!

So, regarding the duplication, you are right.
It's a usual problem with xtext that is not easily to solve, at least not
in the way we would like.
As all this classes like WNamedObject and WClass, are generated based on
the grammar (WollokDsl.xtext),
we don't have full control on them.
XText does a pretty good job on abstracting superclasses and interfaces,
and moving up some fields/methods
but sometimes it doesn't, like in this case.
And we end up with two classes defining the same thing.

One way to overcome that and make them polymorphic is to use extension
methods and multiple dispatch
for example:


def dispatch getTheMixins(WClass c) { return c.mixins }
def dispatch getTheMixins(WNamedObject wko) { return wko.mixins }


Then you can do x.getTheMixins() in both WClass and WNamedObject.

I believe I already wrote this methods in one of our "extension classes".
Look for W*Extension. Like WMethodContainerExtensions, etc. (we could
better organize this extensions, we have a lot).

Anyway, I think we are good for the PR right now.
If you want to refactor this later we can do another.

Thanks !

PR #642, opened 2016-04-17
change Double to BigDecimal

Juancete:
Ok, 

Basically WDouble is now BigDecimal. That solves the problem of multiplication, etc but... (_allways have a but_)..... BigDecimal have no fractional pow(). The only way i found to fix this without an external library is to use Math.pow(a,b) and convert BigDecimal to double :-1:

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok/lang/WDouble.xtend
+++ b/org.uqbar.project.wollok.lib/src/wollok/lang/WDouble.xtend
@@ -4,54 +4,55 @@ import org.uqbar.project.wollok.interpreter.WollokInterpreter
 import org.uqbar.project.wollok.interpreter.WollokRuntimeException
 import org.uqbar.project.wollok.interpreter.core.WollokObject
 import org.uqbar.project.wollok.interpreter.nativeobj.NativeMessage
+import java.math.BigDecimal
 
 /**
  * @author jfernandes
  */
 //TODO: has duplicated code with WInteger
-class WDouble extends WNumber<Double> implements Comparable<WDouble> {
+class WDouble extends WNumber<BigDecimal> implements Comparable<WDouble> {
 	
 	new(WollokObject obj, WollokInterpreter interpreter) {
 		super(obj, interpreter)
 	}
 	
-	def abs() { Math.abs(this.wrapped).asWollokObject }
+	def abs() { this.wrapped.abs.asWollokObject }
 
 	@NativeMessage("+")
 	def plus(WollokObject other) { operate(other) [ doPlus(it) ] }
-		def dispatch Number doPlus(Integer w) { wrapped + w }
-		def dispatch Number doPlus(Double w) { wrapped + w }
+		def dispatch Number doPlus(Integer w) { wrapped + new BigDecimal(w) }
+		def dispatch Number doPlus(BigDecimal w) { wrapped + w }
 		//TODO: here it should throw a 100% wollok exception class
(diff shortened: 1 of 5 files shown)


org.uqbar.project.wollok.lib/src/wollok/lang/WDouble.xtend
```diff
@@ -4,54 +4,55 @@ import org.uqbar.project.wollok.interpreter.WollokInterpreter
...
-		def dispatch Number doDivide(Double w) { wrapped / w }
+		def dispatch Number doDivide(Integer w) { wrapped / new BigDecimal(w) }
+		def dispatch Number doDivide(BigDecimal w) { wrapped / w }
 		def dispatch Number doDivide(Object w) { throw new WollokRuntimeException("Cannot divide " + w) }
 
 	@NativeMessage("**")
 	def raise(WollokObject other) { operate(other) [ doRaise(it) ] }
 		def dispatch Number doRaise(Integer w) { wrapped ** w }
-		def dispatch Number doRaise(Double w) { wrapped ** w }
+		def dispatch Number doRaise(BigDecimal w) { Math.pow( wrapped.doubleValue , w.doubleValue) }
```
npasserini: Why use Math.pow instead of the pow in the bigdecimal object?

I think
```wrapped.pow(w) ```

should be enough.

File: org.uqbar.project.wollok.lib/src/wollok/lang/WInteger.xtend [hunk 2]
```diff
@@ -21,38 +22,38 @@ class WInteger extends WNumber<Integer> implements Comparable<WInteger> {
...
 		def dispatch Number doDivide(Integer w) { wrapped / w }
-		def dispatch Number doDivide(Double w) { wrapped / w }
+		def dispatch Number doDivide(BigDecimal w) { new BigDecimal(wrapped) / w }
 		def dispatch Number doDivide(Object w) { throw new WollokRuntimeException("Cannot divide " + w) }
 		
 	@NativeMessage("**")
 	def raise(WollokObject other) { operate(other) [ doRaise(it) ] }
 		def dispatch Number doRaise(Integer w) { (wrapped ** w).intValue }
-		def dispatch Number doRaise(Double w) { wrapped ** w }
+		def dispatch Number doRaise(BigDecimal w) { Math.pow(wrapped, w.doubleValue) }
```
npasserini: Again, I think we should use bigdecimal pow:

```new BigDecimal(wrapped).pow(w)```

npasserini:
:+1:  Nice!

Just one more question, what about converting numbers to strings, how does it work? do we have a test case for that?

org.uqbar.project.wollok.lib/src/wollok/lang/WDouble.xtend
Juancete: Because wrapped.pow(w) only acepts an Integer w. So, other test fails like assert.equals(9 ** 0.5,3).
The bes way to do this without an external library is using Math.pow()

org.uqbar.project.wollok.lib/src/wollok/lang/WInteger.xtend
Juancete: Again, the same comment :smile:

Juancete:
I think does not exist...

tesonep:
I have a question, in many places the WDouble was converted to Double, so it can be used against the Java API, do you take that in account? I don't know if there is a test of it.

Juancete:
Nope. I think now WDouble operates as a BigDecimal but compares as Double. That´s why test works.

Issue #641, opened 2016-04-13
Duplicate elements in List don't get printed correctly


mmatos:
I did a few queries in the console, and everytime I reached the same bug

>> [[3,5,2], [4,2,6]].flatten()
[3, 5, 2, 4, a Integer, 6]

>> [1,2,3].flatMap({x => [x, x*2, x*3]})
[1, 2, 3, a Integer, 4, 6, a Integer, a Integer, 9]

>> const x = [1,2,3]
[1, 2, 3]
>> x.add(5)
>> x
[1, 2, 3, 5]
>> x.addAll([4,3,5])
>> x
[1, 2, 3, 5, 4, a Integer, a Integer]

>> [1,1,1]
[1, a Integer, a Integer]

PR #650, opened 2016-05-08
Dev 1.4 #641 printing duplicated elements in list

tesonep:
I have fixed when showing duplicated numbers in a List. It is the same behavior when an object has two instances variables with the same numeric value.


javierfernandes:
This only fixes numbers, right ?
This was caused by the mechanism to avoid infinity loops while printing a recursive object graph.

tesonep:
Yes, i have only changed the numbers side. I have changed the base case for the numbers.

javierfernandes:
Cool

Issue #608, opened 2016-03-07
Warnings when running the command line REPL


flbulgarelli:
Hi!

I am getting the following warnings when running the REPL:
```
Warning: NLS unused message: CheckGroup_DEFAULT_GROUP in: org.uqbar.project.wollok.messages
Warning: NLS unused message: CheckGroup_CODE_STYLE in: org.uqbar.project.wollok.messages
Warning: NLS unused message: CheckGroup_POTENTIAL_PROGRAMMING_PROBLEM in: org.uqbar.project.wollok.messages
``` 

Is there any way of hiding them?

Issue #335, opened 2015-10-17
Create categories for grouping validator's checks


javierfernandes:
We have a generic declarative mechanism for validator's checks.
This automatically creates the preferences view.

I would like that to include a way to define the check "@Category".
So that the preferences page will group them instead of having a really big list of fields

Something like this

<img src="https://eclipse.org/Xtext/images/releasenotes/2_8_xtext-validation-preferences.png"/>

javierfernandes:
Implemented !

Looks like this

<img width="856" alt="screen shot 2016-01-02 at 16 16 45" src="https://cloud.githubusercontent.com/assets/4428120/12075840/2b51e27c-b16d-11e5-8923-33e21147fd25.png">

Still we might need to define a couple more categories and categorize the checks.
Although the mechanism is generic and already in place we just need to add an annotation to the check method

```xtend
	@Check
	@DefaultSeverity(ERROR) 
	@CheckGroup(WollokCheckGroup.CODE_STYLE)
	def referenciableNameMustStartWithLowerCase(WReferenciable c) {
             ...
        }
```
The new annotation is **@CheckGroup**. The UI will automatically show it.

PR #651, opened 2016-05-08
Removing unused messages

tesonep:
Fixes issue #608

### DIFF
--- a/org.uqbar.project.wollok/src/org/uqbar/project/wollok/Messages.java
+++ b/org.uqbar.project.wollok/src/org/uqbar/project/wollok/Messages.java
@@ -72,6 +72,14 @@ public class Messages extends NLS {
 	public static String CheckSeverity_INFO;
 	
 	// ****************************
+	// ** Check groups
+	// ****************************
+
+	public static String CheckGroup_DEFAULT_GROUP;
+	public static String CheckGroup_CODE_STYLE;
+	public static String CheckGroup_POTENTIAL_PROGRAMMING_PROBLEM;
+	
+	// ****************************
 	// ** runtime interpreter
 	// ****************************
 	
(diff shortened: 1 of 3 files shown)


javierfernandes:
I think this messages are being used. Because the validator's section in the configuration is now being show into "sections" to group similar rules together.
Probably just the comment got outdated, right ?

Also in the commit I see that you added them, not removed, right ?

tesonep:
I have check if the messages are used, but they are not. They were in the properties but not in the Java side. I comment them because maybe they are used.

javierfernandes:
There are some cases where we use reflection to resolve properties :P
This was to build like a more declarative mini-fwk for validations with configurations

Here is the annotation used to declare that a given validator method belongs to a group

https://github.com/uqbar-project/wollok/blob/master/org.uqbar.project.wollok/src/org/uqbar/project/wollok/validation/WollokCheckGroup.xtend

Here is a sample validator for "code style"
https://github.com/uqbar-project/wollok/blob/master/org.uqbar.project.wollok/src/org/uqbar/project/wollok/validation/WollokDslValidator.xtend#L117

Somewhere, I supposed in the "ui" part there's the preference pages shows the list of checkboxes grouped in panels based on these groups and uses the i18n for showing their names.

javierfernandes:
Ok, here is the issue with a screen shot

https://github.com/uqbar-project/wollok/issues/335

And the line where it resolves the group name in i18n based on a convention

https://github.com/uqbar-project/wollok/commit/2e304df9ddb6e5e2ef796d6f90f3a8683dc5f42e#diff-110562837e0fe4917f38d4c9c2092fccR74

So, seems like it is used

tesonep:
I put back the messages, but I have added the entry in the Java side so the warning goes away. It seems this thing of the PR and reviewing the code is working, these hippies are right.

javierfernandes:
👍

javierfernandes:
Should we merge this ?
Was 1.4.3 released already ?

tesonep:
Yes, the idea is to integrate these changes in 1.4.3 and then in 1.5.0. The 1.4.3 is still open.

Issue #691, opened 2016-06-08
No-args sum, max and min versions 


flbulgarelli:
It would be nice to expose `sum`, `max` and `min` messages in numeric collections, similar to their existing counterparts but that don't require a block. For example:  `[20, 30, 4].sum() //answers 54`

matifreyre:
Easy enough... let me try to do this one.

matifreyre:
@flbulgarelli please have a look at branch dev-issue-#691 and tell me if it looks as you expected.

matifreyre:
Methods added in #718.

fdodino:
@matifreyre is this ready? Can we close this issue?

fdodino:
Sorry, didn't see it was already closed!!

matifreyre:
You can re-open and close it again if you want!

PR #718, opened 2016-06-08
Add methods sum(), min() and max() to collections

matifreyre:
Implements methods requested on #691.

### DIFF
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/CollectionTestCase.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/CollectionTestCase.xtend
@@ -31,6 +31,8 @@ class CollectionTestCase extends AbstractWollokInterpreterTestCase {
 		program p {
 			«instantiateCollectionAsNumbersVariable»		
 			assert.equals(2, numbers.min() )
+			assert.equals(1, [1].min())
+			assert.throwsException({[].min()})
 		}'''.interpretPropagatingErrors
 	}
 		
@@ -53,6 +55,8 @@ class CollectionTestCase extends AbstractWollokInterpreterTestCase {
 		program p {
 			«instantiateCollectionAsNumbersVariable»		
 			assert.equals(22, numbers.max() )
+			assert.equals(1, [1].max())
+			assert.throwsException({[].max()})
 		}'''.interpretPropagatingErrors
 	}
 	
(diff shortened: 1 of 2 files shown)


flbulgarelli:
Looks good to me!

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/CollectionTestCase.xtend
```diff
@@ -296,4 +314,13 @@ class CollectionTestCase extends AbstractWollokInterpreterTestCase {
 		}'''.interpretPropagatingErrors
 	}
 	
+	@Test
+	def void sumNoArgs() {
+		'''
+		program p {
+			«instantiateCollectionAsNumbersVariable»
+			assert.equals(34, numbers.sum())
```
flbulgarelli: @matifreyre  Could we add tests for the edge case of empty and unitary collections?

matifreyre: We certainly can, but now that you mention it... min(closure) and
max(closure), which are the base for min() and max() respectively, use
absolute(closure, criteria) as their base. This method doesn't fail on
empty lists, but returns null. I'd rather have min and max (either version
of each one) fail, what do you think?

The case of sum() or sum(closure) is much simpler, of course.

flbulgarelli: Agree!

matifreyre:
I think this one can be merged... any objections?

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/CollectionTestCase.xtend
```diff
@@ -53,6 +55,8 @@ class CollectionTestCase extends AbstractWollokInterpreterTestCase {
 		program p {
 			«instantiateCollectionAsNumbersVariable»		
 			assert.equals(22, numbers.max() )
+			assert.equals(1, [1].max())
+			assert.throwsException({[].max()})
 		}'''.interpretPropagatingErrors
```
flbulgarelli: As a suggestion for the future, it is better to add more tests for different scenarios than adding more assertion in the same test.

Issue #733, opened 2016-06-12
Add Random Number Generator to Range/List/Sets


vpennella:
The method >>randomBetween(start, end) shoudn't be in the Object class. 
I mean we shoudn't be able to do something like:
2.randomBetween(1, 55)
true.randomBetween(3,5)
"fafas".randomBetween(3,9)

We need to define a method >>atRandom()  in Range, List and in Set as well.
Examples
(1..5).atRandom()
4
[1..10].atRandom()
6
#{3..9}.atRandom()
5

fdodino:
Agree, I would do

2.randomBetween(9)  -- ok, only for Numbers (integers and decimals)
true.randomBetween(9) -- not allowed

and I think random() is better than atRandom(). What do you think @vpennella ?

flbulgarelli:
I personally find `2.randomBetween(6)` weird. It looks like a question, and after a between I expect 2 arguments. 

Sending that message to a range or collection has a lot more meaning to me, but it is also a simpler interfaces: with just a mesage you resolve the problem for all the scenarios.

`atRandom` and `random` are both gramatically correct. The first one is also used un smalltalk. Ruby uses the term `sample` - `[1, 5, 4].sample`  - which is a statistics term.

I prefer the first and the latter since they are adverbs and verbs, instead of the second which is just a noun and means

fdodino:
Ok, sounds good only to have a random / atRanndom only in Range.

vpennella:
Yes, I agree. I think random() sounds good!.

Also, I am seeing in the MonadRandom in Haskell we have the getRandom, 
https://hackage.haskell.org/package/MonadRandom-0.1.6/docs/Control-Monad-Random-Class.html

So for me both  random() / getRandom()  are good names.

fdodino:
In Progress, but some decisions are pending...

PR #740, opened 2016-06-14
Dev range api

### LINKED ISSUE #730: Range doesn't have any, all, fold methods of Collection
It would be very useful to use them, like in prime() definition of Wollok Integers.
Otherwise I have to code

```java
		method prime() {
			if (self == 1) return false
			var isPrime = true
			(2..self - 1).forEach({ i =>
				// Horrible definition, but return doesn't exit from loop
				// and I need fold, any methods in Range
				if(self % i == 0) isPrime = false 
			})
			return isPrime
		}
```

fdodino:
Adds randomUpTo(other) to integer and real numbers.
Deletes randomBetween(min, max) in Wollok object.
Adds new methods in Range, polymorphic to  Collection.
Adds new javadoc comments in Integer and Range.
Fixes #730, #733 issues


fdodino:
I made a mistake: master branch!! So I'll generate a new PR.

PR #741, opened 2016-06-15
Dev range api

### LINKED ISSUE #730: Range doesn't have any, all, fold methods of Collection

fdodino:
Adds randomUpTo(other) to integer and real numbers.
Deletes randomBetween(min, max) in Wollok object.
Adds new methods in Range, polymorphic to Collection.
Adds new javadoc comments in Integer and Range.
Fixes #730, #733 issues

This PR replaces old https://github.com/uqbar-project/wollok/pull/740, where base branch was erroneously master.

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -535,7 +535,7 @@ package lang {
 		method /(other) native
 		method div(other) native
 		/**
-		 * high
+		 * raisedTo
 		 * 3 ** 2 = 9
 		 */
 		method **(other) native
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -710,6 +710,17 @@ package lang {
 			return l
 		}
 		
+		method toList() {
+			return self.map({ elem => return elem })
+		}
+		method fold(seed, foldClosure) { return self.toList().fold(seed, foldClosure) }
+		
+		method size() { return end - start + 1 }
+		
+		method any(closure) { return self.toList().any(closure) }
+		
+		method anyOne() { return self.toList().anyOne() }
+		
 		override method internalToSmartString(alreadyShown) = start.toString() + ".." + end.toString()
 	}
 	
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -691,13 +691,21 @@ package lang {
 	class Range {
 		const start
 		const end
+		var step
 		
 		constructor(_start, _end) {
 			self.validate(_start)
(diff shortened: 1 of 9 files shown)


org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -565,15 +588,12 @@ package lang {
...
 		method isPrime() {
 			if (self == 1) return false
-			var _prime = true
-			(2..self - 1).forEach({ i =>
-				// Horrible definition, but return doesn't exit from loop
-				// and I need fold, any methods in Range
-				if(self % i == 0) _prime = false 
-			})
-			return _prime
+			return (2..(self.div(2) + 1)).any({ i => self % i == 0 }).negate()
```
npasserini: Why self.div(2) instead of self / 2?
Or also self.sqrt() or similar.


Also we could replace self % i == 0 by self.isDivisibleBy(i) or i.divides(self)

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -692,14 +714,22 @@ package lang {
...
 			start = _start 
 			end = _end
+			if (_start > _end) { 
+				step = -1 
+			} else {
+				step = 1
+			}  
 		}
 		
+		method step(_step) { step = _step }
```
npasserini: I think ranges should be inmutable. Don't you?

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -710,6 +740,48 @@ package lang {
...
 		
+		/** @private */
+		method asList() {
+			return self.map({ elem => return elem })
+		}
+		
+		method isEmpty() = self.size() == 0
+
+		method fold(seed, foldClosure) { return self.asList().fold(seed, foldClosure) }
+		method size() { return end - start + 1 }
```
npasserini: size does not take the step into account.

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -710,6 +740,48 @@ package lang {
...
+		method any(closure) { return self.asList().any(closure) }
+		method all(closure) { return self.asList().all(closure) }
+		method filter(closure) { return self.asList().filter(closure) }
+		method min() { return self.asList().min() }
+		method max() { return self.asList().max() }
+		/**
```
npasserini: Could be implement all this as a mixin?
That would be nice.

Issue #742, opened 2016-06-15
Add removeAll(anotherCollection) to collections


matifreyre:
The only doubt I have is if it should fail if any of the elements in anotherCollection doesn't exist in the receiver. I see it more useful if it doesn't.
Related to #726.

npasserini:
Just follow the same behavior as in single remove()

Issue #729, opened 2016-06-09
Add filter with effect message


flbulgarelli:
It would be nice to have an equivalent of the `removeAllSuchThat` message of Smalltalk, which is sometimes used in exercises.

Issue #728, opened 2016-06-09
Add collection message to count occurences


flbulgarelli:
It could be named `ocurrencesOf` like Smalltalk's. 

Not really a high priority since it can be implemented in terms of `count`

matifreyre:
Done in dev-collections-api branch.

Issue #727, opened 2016-06-09
Add last method to list


flbulgarelli:
Because it is useful :stuck_out_tongue:

matifreyre:
Done in dev-collections-api branch.

flbulgarelli:
Hi @matifreyre, notice that you can always reference the issue in the commit using the `Fixes #IssueNumer` syntax as part of the comment. 

That makes the issue tracking easier.

PR #744, opened 2016-06-15
Dev collections api

matifreyre:
Addresses several requests of new methods for collections:
- #742 removeAll(another)
- #726 union(another), intersection(another) and difference(another) **for Sets only**
- #729 removeAllSuchThat(closure)
- #728 occurrencesOf(element)
- #727 last()
- Adds doc to several methods and fixes several other methods' doc too.


org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -357,7 +382,14 @@ package lang {
 			acc
 		})
 
-		method filter(closure) = self.fold(self.newInstance(), { acc, e =>
+		/**
+		 * Returns a new collection that contains the elements that meet a given condition.
+		 * The condition is a closure argument that takes a single element and returns a boolean.
+		 * @returns another collection (same type as self one)
```
flbulgarelli: I tend to use `element` for collection members, `argument` or `parameter` - depending context - for closure inputs and `components` for tuple members.

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -407,6 +464,26 @@ package lang {
 		override method asSet() = self
 
 		override method anyOne() native
+
+		/**
+		 * Returns a new Set with the elements of both self and another collection.
+		 * @returns a Set
+		 */
+		 method union(another) = self + another
```
flbulgarelli: Sorry, where is  + defined?

I see `union`, `difference` , etc documented as if they return a set, but it seems they are not.

```
method intersection(another) = self.filter({it => another.contains(it)}) 
```

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/SetTestCase.xtend
```diff
@@ -0,0 +1,83 @@
...
+	@Test
+	def void unionWithNonEmptySets() {
+		'''
+		program p {
+			const aSet = #{1, 2, 3}
+			const anotherSet = #{3, 4, 5}
+			const unionSet = #{1, 2, 3, 4, 5}
+			
+			assert.equals(aSet, aSet.union(aSet))
+			assert.equals(unionSet, aSet.union(anotherSet))
```
flbulgarelli: Could we perhaps define separate tests for separate assertions?

org.uqbar.project.wollok.lib/src/wollok.wlk
matifreyre: + is defined in Collection and returns a collection of the same class as
the receiver (Set or subclass).

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/SetTestCase.xtend
matifreyre: 1 test = 1 assertion? I like it more too, but I see it a bit extreme
sometimes.
Knowing that the whole project already has a different approach, I rather
be consistent with the rest of the project.
Of course we can create a request to refactor the whole project tests
anytime.

org.uqbar.project.wollok.lib/src/wollok.wlk
matifreyre: I'm sorry, I must have messed something up because I don't see it in my branch. I don't really understand how it came to this, I have it locally in my PC because it was implemented in the wdate branch recently, but it doesn't show up in here, not even as deleted.

Now I'll have to review everything to check if it's the only thing missing :(

flbulgarelli: ;)

flbulgarelli: @matifreyre so I have a doubt here: it `union` a synonym of `+` ? Is such definition overriden in any of the collection subclasses?

matifreyre: Yes, it's a synonym.
So far Set doesn't have built-in subclasses, so the answer is that it's not overriden yet. It might be in the future.

npasserini: I am not sure but maybe + is also implemented for lists as concatenation?

matifreyre: It's implemented in Collection, but yes it works as concatenation for lists.

```
		/**
		  * Concatenates all elements from the given collection parameter to self collection giving a new collection
		  * (no side effect) 
		  */
		method +(elements) {
			const newCol = self.copy() 
			newCol.addAll(elements)
			return newCol 
		}
```

flbulgarelli: > I am not sure but maybe + is also implemented for lists as concatenation?

That means that you can say to a list to union another list, and you will get the concatenation of both lists, right? That is weird.

> Yes, it's a synonym.
Ok. I believe that synonyms are a big headache for maintainers - _which method should I override?_ - and students - _which methods should I use?_.

Actually, in another issue I am requesting to remove the equal/== synonym.  

What about just defining `+` in collections, but `union` just in sets? If we are going to the _union just works for sets_ approach, naming a concatenation as `union` is not a good idea.

matifreyre: > What about just defining + in collections, but union just in sets?

That's what's done. They are synonyms in Set, not in Collection.

flbulgarelli: (Y)

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/SetTestCase.xtend
fdodino: :+1:

Issue #738, opened 2016-06-14
Model an Interval object for real (double, float) numbers


fdodino:
It should have start, end, and step.
It should offer this methods:
- contains(e)
- random()

It is not enumerable since they are infinite numbers.
You can not implement a forEach method.

npasserini:
Why step?
I see an interval as a continous set of (real) numbers.

tesonep:
It should be in the Math namespace, once the package is fixed.

fdodino:
Ok, no step since it doesn't make sense if not enumerable.
Didn't understand what "Math namespace" is.

tesonep:
I was just saying to put it in the wollok.math package, or something like that, nothing really important.

Issue #670, opened 2016-06-05
Range only for int


JuanchiRios:
Wollok only supports range for integer values?
`class Range extends AbstractJavaWrapper<IntegerRange> `

If i do
```
var rango =new Range(5,10)
rango.map({x=>x*2}).toString()
```
i get [10, 12, 14, 16, 18, 20]

but if i do

```
var rango = new Range(5.0,10.0)
rango.map({x=>x*2.0}).toString()
```
i get
`wollok.lang.Exception: java.math.BigDecimal cannot be cast to java.lang.Integer
	at wollok.lang.Range.forEach(closure) [/wollok.wlk:602]
	at wollok.lang.Range.map(closure) [/wollok.wlk:606]
	at  [/test.wtest:5]`

Why Wollok allow me to construct a range of doubles if i cannot do anything with it? My point here is being consistent, don't allow me to construct an invalid range of Doubles values or  let me make range of nums to play with them.

fdodino:
I added a native validate() method. Push pending.

fdodino:
Solved in wdate branch

PR #752, opened 2016-06-18
Interval class for a set of continuous (real) numbers

fdodino:
See issue #738 
Also fixed a copy/paste problem in WollokJavaConversions for double.

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok/lang/Interval.xtend
+++ b/org.uqbar.project.wollok.lib/src/wollok/lang/Interval.xtend
@@ -16,16 +16,18 @@ class Interval {
 	def solve(String fieldName) {
 		obj.resolve(fieldName).wollokToJava(Double) as BigDecimal
 	}
-	
+
+	/**
+	 * Type constraint: only numbers allowed, integers and doubles
+	 */	
 	def void validate(WollokObject value) {
 		try {
 			value.wollokToJava(Double) as BigDecimal
 		} catch (Throwable e) {
 			try {
-				// Permitimos casteo de entero a BigDecimal
 				value.wollokToJava(Integer) as BigDecimal
 			} catch (Throwable eInt) {
-				throw new IllegalArgumentException(value.toString() + " : only reals are allowed in an Interval")
+				throw new IllegalArgumentException(value.toString() + " : only numbers are allowed in an Interval")
 			}
 		}
 	}
--- a/org.uqbar.project.wollok.lib/src/wollok/lang/Interval.xtend
+++ b/org.uqbar.project.wollok.lib/src/wollok/lang/Interval.xtend
@@ -1,0 +1,35 @@
+package wollok.lang
+
+import org.uqbar.project.wollok.interpreter.WollokInterpreter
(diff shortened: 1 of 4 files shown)


tesonep:
Hello Fer, why do we have the validate to checking the parameters? Why not directly using them??
And also we can not implement the solution all in wollok??

fdodino:
>Hello Fer, why do we have the validate to checking the parameters? Why not directly using them??
Well, there was an issue reported:
https://github.com/uqbar-project/wollok/issues/670
So I implemented a check in Range and also here, casting to Double values. 
Otherwise if you try to use an Interval with Strings, it will fail after initialization.

> And also we can not implement the solution all in wollok??
Is there a random object or class?

org.uqbar.project.wollok.lib/src/wollok/lang/Interval.xtend
```diff
@@ -0,0 +1,39 @@
...
+	
+	def void validate(WollokObject value) {
+		try {
+			value.wollokToJava(Double) as BigDecimal
+		} catch (Throwable e) {
+			try {
+				// Permitimos casteo de entero a BigDecimal
+				value.wollokToJava(Integer) as BigDecimal
+			} catch (Throwable eInt) {
+				throw new IllegalArgumentException(value.toString() + " : only reals are allowed in an Interval")
```
npasserini: Given that we accept integers, I would change the comment from "real" to "number"

npasserini:
>> And also we can not implement the solution all in wollok??
> Is there a random object or class?

There is no random object, so that part has to be done natively. I am not sure about adding one random object, should we discuss that?

> Seemed easier to implement it in Xtend.

"Easier" is kind of not enough :P
I would not get crazy about it, it is just that we prefer to implement in Wollok everything that is "possible".
I am not sure if it is posible to "go more native"... just commenting an idea.

> Hello Fer, why do we have the validate to checking the parameters? Why not directly using them??
Given that it is a library that will be used by novice students, I think we should maximize the validations because this allows us to detect mis-usage, and provide better feedback.

npasserini:
My only issue with this pull request is that I am not sure that this object is necessary, it just came accross in a discussion. But maybe I am being just too conservative. 

Also it is not a big deal, just that I would not like to make the language bigger without a good reason, I like the simplicity and sort of minimality of current core classes.

Do you think that it will be frequently used for teaching examples?
If so just merge it.

org.uqbar.project.wollok.lib/src/wollok/lang/Interval.xtend
fdodino: :+1:

fdodino:
>Also it is not a big deal, just that I would not like to make the language bigger without a good reason, I like the simplicity and sort of minimality of current core classes.

Didn't know about this principle.
I suppose I don't have to worry to have a LocalTime object  :)

fdodino:
Main requirement is to have a random object for certain exercises. Very often they appear in initial courses. Less important is the concept of an interval object. :+1: if we have it.

PR #753, opened 2016-06-20
Dictionary/Map support issue #387

fdodino:
Developed a new branch from dev (doesn't have test refactor for testing wollok classes, pull request pending for branch dev-test-easy).
Still doesn't support literals but has almost all methods from Map & Dictionary intersection (not union :))

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -572,7 +572,6 @@ package lang {
 		 * 
 		 */
 		method forEach(closure) native
-		method newInstance() = new Dictionary()
 		
 		/**
 		 * Removes all of the mappings from this Dictionary. This is a side-effect operation.
--- a/org.uqbar.project.wollok.lib/src/wollok.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok.wlk
@@ -509,44 +509,60 @@ package lang {
 		method put(_key, _value) native
 		
 		/*
-		 * Returns the value to which the specified key is mapped, or null if this map contains no mapping for the key.
+		 * Returns the value to which the specified key is mapped, or null if this Dictionary contains no mapping for the key.
 		 */
-		method get(_key) native
+		method basicGet(_key) native
+
+		/*
+		 * Returns the value to which the specified key is mapped, or evaluates a non-parameter closure otherwise 
+		 */
+		method getOrElse(_key, _closure) {
+			const value = self.basicGet(_key)
+			if (value == null) 
+				_closure.apply()
+			else 
+				return value
+		}
 		
+		/*
+		 * Returns the value to which the specified key is mapped. If this Dictionary contains no mapping for the key, an error is thrown.
+		 */
+		method get(_key) = self.getOrElse(_key,{ => throw new ElementNotFoundException("there is no element associated with key " + _key) })
+
 		/**
-		 * Returns the number of key-value mappings in this map.
+		 * Returns the number of key-value mappings in this Dictionary.
 		 */
 		method size() = self.values().size()
 		
 		method isEmpty() = self.size() == 0
 		
 		/**
-		 * Returns true if this map contains a mapping for the specified key.
+		 * Returns true if this Dictionary contains a mapping for the specified key.
 		 */
 		method containsKey(_key) = self.keys().contains(_key)
 		
 		/**
-		 * Returns true if this map maps one or more keys to the specified value.
+		 * Returns true if this Dictionary maps one or more keys to the specified value.
 		 */
 		method containsValue(_value) = self.values().contains(_value)
 		
 		/**
-		 * Removes the mapping for a key from this map if it is present 
+		 * Removes the mapping for a key from this Dictionary if it is present 
 		 */
 		method remove(_key) native
 		
 		/**
-		 * Returns a list of the keys contained in this map.
+		 * Returns a list of the keys contained in this Dictionary.
 		 */
 		method keys() native
 		
 		/**
-		 * Returns a list of the values contained in this map.
+		 * Returns a list of the values contained in this Dictionary.
 		 */
 		method values() native
 		
 		/**
-		 * Performs the given action for each entry in this map until all entries have been 
+		 * Performs the given action for each entry in this Dictionary until all entries have been 
 		 * processed or the action throws an exception.
 		 * 
 		 * Expected closure with two parameters: the first associated with key and second with value.
@@ -559,7 +575,7 @@ package lang {
 		method newInstance() = new Dictionary()
 		
 		/**
-		 * Removes all of the mappings from this map. This is a side-effect operation
+		 * Removes all of the mappings from this Dictionary. This is a side-effect operation.
 		 */
 		method clear() native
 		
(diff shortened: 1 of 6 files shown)


org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -492,6 +496,24 @@ package lang {
...
+		method containsKey(_key) = return self.keys().contains(_key)
+		method containsValue(_value) = return self.values().contains(_value)
+		method remove(_key) native
+		method keys() native
+		method values() native
+		method forEach(closure) native
+		method newInstance() = return new Dictionary()
+		method clear() native
+		
+	}
```
matifreyre: I know that it doesn't hurt anyone, but shortcut methods don't need the ```return``` keyword.

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -492,6 +496,24 @@ package lang {
...
+		method put(_key, _value) native
+		method get(_key) native
+		method size() = return self.values().size()
+		method isEmpty() = return self.size() == 0
+		method containsKey(_key) = return self.keys().contains(_key)
+		method containsValue(_value) = return self.values().contains(_value)
+		method remove(_key) native
+		method keys() native
+		method values() native
+		method forEach(closure) native
```
matifreyre: I'd have some doc added to ```forEach```. It's not clear if ```closure``` works on keys, values or both.

fdodino: :+1:

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -105,7 +105,11 @@ package lang {
 		method !=(other) = ! (self == other)
 		
 		/**
-		 * Tells whether self object is identical (the same) to the given one.
+		 *
+		 
+		 
+		 
+		  Tells whether self object is identical (the same) to the given one.
```
matifreyre: A really minor detail... what's with this multi-blank-line comment?

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -493,6 +493,91 @@ package lang {
...
+	class Dictionary {
+	
+		constructor() { }
+		/**
+		 * Adds or updates a value based on a key
+		 */
+		method put(_key, _value) native
+		
+		/*
+		 * Returns the value to which the specified key is mapped, or null if this Dictionary contains no mapping for the key.
```
npasserini: This implies that the dictionary can not differentiate not having a key from associating a null value with that key. I do not say it is wrong, just bring it up to be sure that it is what we mean.

npasserini: Nevertheless, the underlying java HashMap allows for null values, so there might be inconsistencies.

matifreyre: Fer already changed it, although he may not have pushed it yet. Now there
is a basicGet (like this one) a standard get, which throws an exception
when the key is not found, and a getOrElse that lets you choose the
alternative using a block.

org.uqbar.project.wollok.lib/src/wollok.wlk
```diff
@@ -493,6 +493,91 @@ package lang {
...
+		 * processed or the action throws an exception.
+		 * 
+		 * Expected closure with two parameters: the first associated with key and second with value.
+		 *
+		 * Example:
+		 * 		mapaTelefonos.forEach({ k, v => result += k.size() + v.size() })
+		 * 
+		 */
+		method forEach(closure) native
+		method newInstance() = new Dictionary()
```
npasserini: Do we need this?

org.uqbar.project.wollok.lib/src/wollok/lang/WDictionary.xtend
```diff
@@ -0,0 +1,60 @@
...
+	
+	def basicGet(WollokObject key) { wrapped.get(getInternalKey(key)) }
+	
+	def put(WollokObject key, WollokObject value) { wrapped.put(key, value) }
+
+	def void remove(WollokObject key) {
+		wrapped.remove(getInternalKey(key))
+	}
+	
+	private def Object getInternalKey(WollokObject key) {
```
npasserini: Even when we are not performance maniacs... I think this breaks the idea of having a dictionary, we shouldn't iterate through the keys to find the right object. ¿Is it too difficult to have a hash+equals which are compliant with the needs of the hashmap?

npasserini:
I would like that we (very briefly) discuss the issue about the null values. Do we allow them or not? If we don't, should we check it?

Right now I think that we could break a dictionary like this:

Supose I receive x, which is null, but my program does not have to be aware of it:

doSth(x) {
  ...
  map.put(key, x)
 ...
  map.get(key) => error there is no value for key
}

On the other side there might be a performance issue... which could be considered secondary.


If needed I think that this can be merged, provided that we create issues for both comments.

org.uqbar.project.wollok.lib/src/wollok/lang/WDictionary.xtend
fdodino: I originally used wrapped.get(key), and wrapped.remove(key) but two Strings as WollokObjects are never equals. So get method always returned null.
I was stuck until I saw WList implementation of remove.
I'm not sure if there is a possibility of changing WollokObject definition of equals / hash to be like wollokEquals.

org.uqbar.project.wollok.lib/src/wollok.wlk
fdodino: I don't know, maybe it prepends an algorithmic way of thinking.

npasserini: I do not understand... or maybe it was I that did not ask a meaningful question.
I meant if we need the newInstance method. 

I understand it for list/set because it allows a way to reuse code for methods like map/filter/etc but for Dictionaries I do not see why we would need it.

org.uqbar.project.wollok.lib/src/wollok/lang/WDictionary.xtend
```diff
@@ -26,7 +26,11 @@ class WDictionary implements JavaWrapper<Map> {
 	
 	def basicGet(WollokObject key) { wrapped.get(getInternalKey(key)) }
 	
-	def put(WollokObject key, WollokObject value) { wrapped.put(key, value) }
+	def put(WollokObject key, WollokObject value) {
+		if (key == null) throw new IllegalArgumentException("You cannot put a null key in a Dictionary") 
+		if (value == null) throw new IllegalArgumentException("You cannot put a null value in a Dictionary")
```
npasserini: Maybe it is just my lack of knowledge, but is it right to throw IllegalArgumentException?
Will it be correctly informed to the programmer?
Shouldn't we throw a wollok exception?

fdodino: If you run it in a REPL console this is what you get:

>>> const mapa = new Dictionary()
a Dictionary[]
>>> mapa.put(2, null)
wollok.lang.Exception: You cannot put a null value in a Dictionary
	at wollok.lang.Dictionary.put(_key,_value) [/wollok.wlk:505]
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
	at  [/wollokREPL.wlk:1]

Is it ok?

PR #770, opened 2016-06-27
Dev #765 anonymous objects are not printable

### LINKED ISSUE #765: Anonymous objects are not printable
It would be nice to make objects defined in repl implement `toString`. Otherwise you see this error:

```
>>> [mobject {  var x = 3   }
[31mwollok.lang.MessageNotUnderstoodException: anonymousObject does not understand message toString()
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
        ....
```

PalumboN:
Close #765 

The problem was anonymous objects didn't inherits from Object, so I added WObjectLiteral in the _WollokLinker_.
Then Wollok fails on execute `toString()` method, and it was because the Interpreter didn't add the inherits elements (and its natives). I change the `evaluate(WObjectLiteral l)` method in _WollokInterpreterEvaluator_ and delegate some duplicated code into methods.

### Some unrelated problems 

1. On add `getParentRef(WObjectLiteral wObject)` in the multi-method I get a compilation error: `Type mismatch: cannot convert from EObject to WMethodContainer`. 
I'm not sure why, but I can fix it adding a definition for EObject: `getParentRef(EObject wObject)`. 
But It's sounds to _hack_, there is other solution?

2. Wollok has a bug on printing objects with a Double property type: https://github.com/uqbar-project/wollok/issues/771
I had to change the test for avoid this problem.

### DIFF
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/sdk/ObjectTest.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/sdk/ObjectTest.xtend
@@ -34,7 +34,7 @@ class ObjectTest extends AbstractWollokInterpreterTestCase {
 					var altura = 2
 				}
 				
-				assert.equals("anObject[edad=23, altura=2]", anonymousObject.toString())
+				assert.equals("an Object[edad=23, altura=2]", anonymousObject.toString())
 			}
 		'''.interpretPropagatingErrors
 	}
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/sdk/ObjectTest.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/sdk/ObjectTest.xtend
@@ -31,10 +31,10 @@ class ObjectTest extends AbstractWollokInterpreterTestCase {
 				
 				const anonymousObject = object {
 					var edad = 23
-					var altura = 1.7
+					var altura = 2
 				}
 				
-				assert.equals("anObject[edad=23, altura=1.7]", anonymousObject.toString())
+				assert.equals("anObject[edad=23, altura=2]", anonymousObject.toString())
 			}
 		'''.interpretPropagatingErrors
 	}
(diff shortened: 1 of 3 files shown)


javierfernandes:
Right. There was an issue but I cannot find it anymore about making "anonymous" object also inherit from wollok.lang.Object :P

You see we added wollok.lang.Object after they were already there.
Actually anonymous objects were the first kind of objects in Wollok.

Anyway, you need to implement the getParentRef() and all the logic to hook object into the hierarchy because xtext sometimes cannot infer right the fact that a couple of rules like WObjectLiteral, WNamedObejct, and WCLass they all define the "parent" attribute.
So it generates that getters/setters for each class as if they were different things.
So we lose polymorphism.

THen we need to code a multiple dispatch to overcome that like:

```xtend
def dispatch getParent(WClass it) { parent }
def dispatch getParent(WObjectLiteral it) { parent }
def dispatch getParent(WNamedObejct it) { parent }
```

They look the same but they are not statically.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/sdk/ObjectTest.xtend
```diff
@@ -31,11 +31,10 @@ class ObjectTest extends AbstractWollokInterpreterTestCase {
 				
 				const anonymousObject = object {
 					var edad = 23
-					var altura = 1.7
+					var altura = 2
 				}
 				
-«««				//TODO
-«««				assert.equals("anObject[edad=23, altura=1.7]", anonymousObject.toString())
+				assert.equals("an Object[edad=23, altura=2]", anonymousObject.toString())
```
javierfernandes: Maybe this should be a diferente test method with an specific name like "toStringWithDoubleField()" or something.

PalumboN: Mmm I don't know if I understood. But I don't want an special toString() method for objects with Double fields :P.

Anyway, I was thinking do it in the PR of https://github.com/uqbar-project/wollok/issues/771

PalumboN:
> Anyway, you need to implement the getParentRef() and all the logic to hook object into the hierarchy 

But these classes are not autogenerated? How do I do that?

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/sdk/ObjectTest.xtend
PalumboN: Done! https://github.com/uqbar-project/wollok/pull/775

javierfernandes:
I think that it is what you already did here:

```
def dispatch EReference getParentRef(WObjectLiteral wObject) { WollokDslPackage.Literals.WOBJECT_LITERAL__PARENT }
```

Indeed you cannot touch autogenerated classes. So one needs multimethods to add behavior to many classes polymorphically.

Issue #716, opened 2016-06-08
[Weird behavior] Null is a null object? 


flbulgarelli:
Hi!

I found no reference about this behavior in [the wiki](https://github.com/uqbar-project/wollok/wiki/LanguageReference) , but it seems like adding or substracting nulls actually work. Other arithmetic operation works in unexpected ways, and other messages fails in weird ways. 

Some examples that show the heavy inconsistencies: 

 ```
[36;1mConsola Interactiva de Wollok (escriba "quit" para salir): [m
[36m>>> [m10 + null
[34m10[m
[36m>>> [m10 * null
[34mnull[m
[36m>>> [mnull.toString()
[31mwollok.lang.Exception: Cannot send message toString() to null
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
	at  [/metodos.wlk:2]
[m
[36m>>> [m5.toString()
[34m5[m
[36m>>> [m5.toString().kindName()
[34mun/a  String[m
[36m>>> [mnull.methods()
[31mwollok.lang.Exception: Cannot send message methods() to null
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
	at  [/__synthetic0.wpgm:3]
	at  [/metodos.wlk:2]
[m
[36m>>> [mnull / 3
     [31mError de la VM de Wollok en la linea (1): null / 3:[m
[31mjava.lang.NullPointerException
	at org.uqbar.project.wollok.interpreter.operation.WollokDeclarativeNativeBasicOperations$3.apply(WollokDeclarativeNativeBasicOperations.java:97)
	at org.uqbar.project.wollok.interpreter.operation.WollokDeclarativeNativeBasicOperations$3.apply(WollokDeclarativeNativeBasicOperations.java:1)
	at org.uqbar.project.wollok.interpreter.WollokInterpreterEvaluator._evaluate(WollokInterpreterEvaluator.java:717)
	at org.uqbar.project.wollok.launch.WollokLauncherInterpreterEvaluator.evaluate(WollokLauncherInterpreterEvaluator.java:168)
	at org.uqbar.project.wollok.interpreter.WollokInterpreterEvaluator.evaluate(WollokInterpreterEvaluator.java:1)
[...]

Issue #705, opened 2016-06-08
Equals or == ?


flbulgarelli:
Exploring the autocomplete suggestions, I see that it offers you both `equal` and `==`. Since I know the implementation, I know that they are synonyms 

But it is quite confusing for the newcomer if you see these two options, plus the`===`. It is not obvious which one you should use. What is more, you could redefine equals instead of `==`, getting strange problems, so they are not truly synonyms. 

Which is the recommended way of comparing objects by equality? Could we just have one way and discard the other, reducing code divergence and potential errors?

matifreyre:
I like to have both options, but I see the point in having only one of them. 
What about ```&&``` / ```and``` and ```||``` / ```or```? Would you rather have only one option there too?

iluetich:
I think it's rarely considered "good" to have methods / logic that behaves identically. In the example that @matifreyre provides, Ruby gives both options: `&&` and `and`.

However, this operators are not _equal_. See [&& versus and operator](http://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby) and [|| versus or operator](http://stackoverflow.com/questions/2083112/difference-between-or-and-in-ruby). For additional documentation, [here](http://phrogz.net/programmingruby/language.html#table_18.4).

Is there any particular need for having `equals` method **and** `==` operator?

fdodino:
@javierfernandes had an explanation: == is an operand, equals is a message. 
I don't mind having both ways, but I'd like to have a better javadoc to avoid confusions.

flbulgarelli:
> What about && / and and || / or? Would you rather have only one option there too?

My opinion there is the same: we should only have one way . But it is not a so big problem since you are not going probably to override `&&` or `||` methods

flbulgarelli:
> @javierfernandes had an explanation: == is an operand, equals is a message. 

This is not actually true. Both are messages, whose methods are defined in object. And you can actually override them. Actually many operands are messages, for example `+`, and this is good! Actually there should not be special operands, everything should be a message unless there is a really good motivation to not do so.

flbulgarelli:
> I don't mind having both ways, but I'd like to have a better javadoc to avoid confusions.

I don't think having a better _wollokdoc_ ( :stuck_out_tongue: ) is the solution to the problem - if your design is not  clear, documentation is not the solution. Look at the following scenario: I want to define a `Point` class and I want to implement equality. Can you guess which of the following implementations are correct and which not? 

```
class Punto {
	const x
[...]

Issue #666, opened 2016-05-28
add expected Exception type to assert.throwsException 


asanzo:
something like:

`assert.throwsException(MyError, { a.method1() })`

asanzo:
BTW, **diabolic** issue.

Juancete:
The solution that was implemented accepts something like this: 

assert.throwsException(new MyError(), { a.method1() })

it's fine?

fdodino:
:+1:

matifreyre:
Does that implementation take into consideration any of the exception's attributes, like the associated text?

fdodino:
I was also needing this:

``` java
assert.throwsException("Cannot send message + to null", { a.method() })
```

So compare message or exception thrown with or without message (optional if it is not present) would be nice!

Juancete:
@fdodino and @matifreyre 
Initially only compares the exception class. My idea will work like this:

`assert.throwsException(new BusinessException(), { => throw new BusinessException() })`

Works fine. Only checks the exception class, but if you do 

`assert.throwsException(new BusinessException("lalala"), { => throw new BusinessException() })`

Fails because an error message is seted and is not equals. It will be fine if you do something like

`assert.throwsException(new BusinessException("lalala"), { => throw new BusinessException("lalala") })`

is that ok?

matifreyre:
I'm more worried about a case such as this:

assert.throwsException(new BusinessException(), { => throw new
BusinessException("lalala") })

Will it succeed?

Juancete:
Yes, it's confusing. In that case only checks the exception type. So it will succeed. Because i think _"I want to test if i get the same exception class, and if i set an error message in the expected object, i'll check it too.."_ 

if both are equals when have the same class and error message then 
```xtend
assert.throwsException(new BusinessException(), { => throw new
BusinessException("lalala") })
```
fails and 
```xtend
assert.throwsException(new BusinessException("lalala"), { => throw new
BusinessException("lalala") }) 
```
it's ok

asanzo:
I like the approach of comparing both type and message, yet i find it unnerving having this succeed:

```xtend
assert.throwsException(new BusinessException(), { => throw new BusinessException("lalala") })
```

Why not just do a **deep equality**? That is quite simple, easy to explain and to understand. And it will draw attention to both Exception type & description. If I have to teach Exceptions to someone who doesn't know a thing about them, I'd go for this for simplicity.

[...]

PR #772, opened 2016-06-27
Dev null fix

fdodino:
Related issue: #716 

See NullTestCases:
``` java
		assert.throwsException({ null.sayHi() })
		assert.throwsException({ null.toString() })
		assert.throwsException({ null + 3 })
		assert.throwsException({ null - 3 })
		assert.throwsException({ null * 3 })
		assert.throwsException({ null / 3 })
		assert.throwsException({ null || true })
		assert.throwsException({ null && true })
```

But I left == and !=:

``` java
		assert.notThat(null == 8)
```

### DIFF
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
@@ -33,7 +33,11 @@ class NullTestCase extends AbstractWollokInterpreterTestCase {
 	@Test
 	def void equalEqualOperatorSentToNull() {
 		'''
+		var valorNulo
 		assert.notThat(null == 8)
+		assert.notThat(valorNulo == 8)
+		assert.that(valorNulo == null)
+		assert.notThat(valorNulo != null)
 		'''.test
 	}
 
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
@@ -36,5 +36,21 @@ class NullTestCase extends AbstractWollokInterpreterTestCase {
 		assert.notThat(null == 8)
 		'''.test
 	}
+
+	@Test
+	def void ifInANullExpression() {
+		'''
+		var expresionNula
+		var mensajeError
+		try {
+			if (expresionNula) {
+				console.println("ooo")
+			}
+		} catch e: Exception {
+			mensajeError = e.getMessage()
+		}
+		assert.equals("Cannot use null in 'if' expression", mensajeError)
+		'''.test
+	}
 	
 }
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
@@ -4,43 +4,96 @@ import org.junit.Test
 
 class NullTestCase extends AbstractWollokInterpreterTestCase {
 	
+	def defineAssertWithMessage() {
+		'''
+		object extendedAssert {
+			method assertException(closure, msg) {
+				try {
+					closure.apply()
+				} catch e: Exception {
+					assert.equals(msg, e.getCause().getMessage())
+				}
+			}
+		}
+		'''
+	}
+	
 	@Test
 	def void messageSentToNull() {
 		'''
-		assert.throwsException({ null.sayHi() })
-		assert.throwsException({ null.toString() })
-		'''.test
+		«defineAssertWithMessage»
+		program a {
+			extendedAssert.assertException({ null.sayHi() }, "Cannot send message sayHi() to null")
+			extendedAssert.assertException({ null.toString() }, "Cannot send message toString() to null")
+		}
+		'''.interpretPropagatingErrors
 	}
 
 	@Test
 	def void numberOperatorsSentToNull() {
 		'''
-		assert.throwsException({ null + 3 })
-		assert.throwsException({ null - 3 })
-		assert.throwsException({ null * 3 })
-		assert.throwsException({ null / 3 })
-		'''.test
+		«defineAssertWithMessage»
+		program a {
+			extendedAssert.assertException({ null + 3 }, "Cannot send message + to null")
+			extendedAssert.assertException({ null - 3 }, "Cannot send message - to null")
+			extendedAssert.assertException({ null * 3 }, "Cannot send message * to null")
+			extendedAssert.assertException({ null / 3 }, "Cannot send message / to null")
+		}
+		'''.interpretPropagatingErrors
 	}
 	
 	@Test
 	def void booleanOperatorsSentToNull() {
 		'''
-		assert.throwsException({ null || true })
-		assert.throwsException({ null && true })
-		'''.test
+		«defineAssertWithMessage»
+		program a {
+			extendedAssert.assertException({ null || null }, "Cannot send message || to null")
+			extendedAssert.assertException({ null && null }, "Cannot send message && to null")
+		}
+		'''.interpretPropagatingErrors
 	}
 	
 	@Test
 	def void equalEqualOperatorSentToNull() {
 		'''
-		var valorNulo
-		assert.notThat(null == 8)
-		assert.notThat(valorNulo == 8)
-		assert.that(valorNulo == null)
-		assert.notThat(valorNulo != null)
-		'''.test
+		class Golondrina { var energia = 0 }
+		
+		program a {
+			var valorNulo
+			assert.notThat(null == 8)
+			assert.notThat(null == "pepe")
+			assert.notThat(null == 3.0)
+			assert.notThat(null == 1..2)
+			assert.notThat(null == [1,2,3])
+			assert.notThat(null == #{1,2,3})
+			assert.notThat(null == assert)
+			assert.notThat(null == new Golondrina())
+			assert.notThat(valorNulo == 8)
+			assert.that(valorNulo == null)
+			assert.notThat(valorNulo != null)
+		}
+		'''.interpretPropagatingErrors
 	}
 
+	@Test
+	def void equalEqualOperatorWithANullArgument() {
+		'''
+		class Golondrina { var energia = 0 }
+		
+		program a {
+			assert.notThat(8 == null)
+			assert.notThat("pepe" == null)
+			assert.notThat(false == null)
+			assert.notThat(2.0 == null)
+			assert.notThat(1..2 == null)
+			assert.notThat([1,2,3] == null)
+			assert.notThat(#{1,2,3} == null)
+			assert.notThat(assert == null)
+			assert.notThat(new Golondrina() == null)
+		}
+		'''.interpretPropagatingErrors
+	}
+	
 	@Test
 	def void ifInANullExpression() {
 		'''
@@ -48,7 +101,7 @@ class NullTestCase extends AbstractWollokInterpreterTestCase {
 		var mensajeError
 		try {
 			if (expresionNula) {
-				console.println("ooo")
+				assert.fail("Null Expression didn't fail!")
 			}
 		} catch e: Exception {
 			mensajeError = e.getMessage()
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
@@ -8,13 +8,32 @@ class NullTestCase extends AbstractWollokInterpreterTestCase {
 	def void messageSentToNull() {
 		'''
 		assert.throwsException({ null.sayHi() })
+		assert.throwsException({ null.toString() })
 		'''.test
 	}
 
 	@Test
-	def void operatorSentToNull() {
+	def void numberOperatorsSentToNull() {
 		'''
 		assert.throwsException({ null + 3 })
+		assert.throwsException({ null - 3 })
+		assert.throwsException({ null * 3 })
+		assert.throwsException({ null / 3 })
+		'''.test
+	}
+	
+	@Test
+	def void booleanOperatorsSentToNull() {
+		'''
+		assert.throwsException({ null || true })
+		assert.throwsException({ null && true })
+		'''.test
+	}
+	
+	@Test
+	def void equalEqualOperatorSentToNull() {
+		'''
+		assert.notThat(null == 8)
 		'''.test
 	}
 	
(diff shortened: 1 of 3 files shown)


flbulgarelli:
Awesome!

Just to be sure:

* which kind of exception do all the described operations throw? The same kind or different kind?
* what does occur when you send synonyms or special message-like constructs `equals`, `and`, `if`? For instance `null and null` or `if(null) { ... }` ? Do they also fail with an exception, @fdodino ?
* what about `toString`? Does it still work, right?

javierfernandes:
if(null)

And thinks like

if (2)

if ("as")

or even

if (a + 2)

can be statically validated by the checker.
Besides any runtime check, let say

flbulgarelli:
> if(null)
> can be statically validated by the checker.
> Besides any runtime check, let say

That is great!

But it is not enough, since there are scenarios that can not be checked statically. For example: 

```
object o1 { method foobar(x) { if (x) { ... }  }
object o2 { method foobar(x) = x

var foo = null
[o1, o2].random().foobar(foo)
```

So my comment was not about validating the scenario where a null is a literal, but the more general scenario where the value is a null reference. It is very important to have consistent behavior in the non-happy paths, because good, surprise-free error helps students to find the problem when the code doesn't work

javierfernandes:
Yeap.

Also a type system could help catching more cases statically.

But well.. yes, the if and any boolean expression component should check
for null

fdodino:
>which kind of exception do all the described operations throw? The same kind or different kind?
As I mentioned in wollok-dev group, I could not create same Exception as in messages. So not for now.
I'll try to find out why, not easy for me.

>what does occur when you send synonyms or special message-like constructs equals, and, if? For >instance null and null or if(null) { ... } ? Do they also fail with an exception, @fdodino ?
var nulo1
if (nulo1) console.println("jaja") else console.println("jojo")
throws WVM Error in line  (1): if (nulo1) console.println("jaja") else ("jojo"):
java.lang.NullPointerException
Since if is not a message, it is natural. I'll try to see where is it defined.

var nulo2
nulo1 and nulo2
throws org.uqbar.project.wollok.interpreter.WollokRuntimeException: Cannot send message and to null

nulo1.equals(nulo2)
throws wollok.lang.Exception: Cannot send message equals(nulo2) to null
It is ok for me.

> what about toString? Does it still work, right?
Did it? It should not. Why should it understand toString?

fdodino:
>As I mentioned in wollok-dev group, I could not create same Exception as in messages. So not for now.
I'll try to find out why, not easy for me.

Good news!
I think I got it!

npasserini:
Just a little comment:

On Mon, Jun 27, 2016 at 9:28 PM, javierfernandes <notifications@github.com>
wrote:

> if (a + 2)


This can't be checked statically because + is a message and therefore you
do not know if it returns a boolean value or not.... unless you have a type
system :)

javierfernandes:
Yeap @npasserini this is karma :S 

Operator overloading :(

Althought we could have it as a particular configurable rule for students. For the moment (?)

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
```diff
@@ -0,0 +1,60 @@
+package org.uqbar.project.wollok.tests.interpreter
+
+import org.junit.Test
+
+class NullTestCase extends AbstractWollokInterpreterTestCase {
+	
+	@Test
+	def void messageSentToNull() {
+		'''
+		assert.throwsException({ null.sayHi() })
```
npasserini: Is it possible to assert that the correct exception is thrown?
(i.e. differentiate "does not understand" from "message to nothing".

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
```diff
@@ -0,0 +1,60 @@
...
+		assert.throwsException({ null || true })
+		assert.throwsException({ null && true })
+		'''.test
+	}
+	
+	@Test
+	def void equalEqualOperatorSentToNull() {
+		'''
+		var valorNulo
+		assert.notThat(null == 8)
```
npasserini: I would like to include also

assert.notThat(8 == null)
and in fact we should test with different kind of receivers.

Crazy idea just from the top of my head, it would be simpler if the interpreter detects hat anyNotNullObject == null is false and avoid sending a message. Maybe that could be the default behavior of == (in object class) and then you only should override equals:

class Object {
 method ==(other) { return other != null && self.equals(other) }
}

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
```diff
@@ -0,0 +1,60 @@
...
+
+	@Test
+	def void ifInANullExpression() {
+		'''
+		var expresionNula
+		var mensajeError
+		try {
+			if (expresionNula) {
+				console.println("ooo")
+			}
```
npasserini: It would be nice to fail here, no?

npasserini:
Implementation looks good for me. I just would love to add those little details on tests, if possible.

Also I was wandering if we could add validations to right hand side operands. Because otherwise 4 + nothing could produce an error which is different from nothing + 4 which is suboptimal (to say the least). 

But I do not have an answer because... since they are messages we should be conservative in asuming that nothing is an invalid value. Also some (all? which?) binary operations are lazy, so we should not evaluate the right hand side until we ensure it is necessary.

fdodino:
Ok, let me finish javadoc issue first and then I'll add some tests.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
matifreyre: It's in progress. See #666.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
fdodino: Added 

```
		method ==(other) {
			return other != null && self === equals(other) 
		}
```

in Object.

fdodino: added lot of comparison in tests. It is tedious to define a single test, so excuse me for this:

``` java
	@Test
	def void equalEqualOperatorSentToNull() {
		'''
		class Golondrina { var energia = 0 }
		
		program a {
			var valorNulo
			assert.notThat(null == 8)
			assert.notThat(null == "pepe")
			assert.notThat(null == 3.0)
			assert.notThat(null == 1..2)
			assert.notThat(null == [1,2,3])
			assert.notThat(null == #{1,2,3})
			assert.notThat(null == assert)
			assert.notThat(null == new Golondrina())
			assert.notThat(valorNulo == 8)
			assert.that(valorNulo == null)
			assert.notThat(valorNulo != null)
		}
		'''.interpretPropagatingErrors
	}

	@Test
	def void equalEqualOperatorWithANullArgument() {
		'''
		class Golondrina { var energia = 0 }
		
		program a {
			assert.notThat(8 == null)
			assert.notThat("pepe" == null)
			assert.notThat(false == null)
			assert.notThat(2.0 == null)
			assert.notThat(1..2 == null)
			assert.notThat([1,2,3] == null)
			assert.notThat(#{1,2,3} == null)
			assert.notThat(assert == null)
			assert.notThat(new Golondrina() == null)
		}
		'''.interpretPropagatingErrors
	}
```

All tests are green

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
fdodino: :+1:

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
fdodino: I made a fix, adding a getCause() method in Exception while @Juancete is finishing #666.
Now you validate "Cannot send message || to null" and other error messages.

fdodino:
Enhanced the tests. Pushing in 3, 2, 1...

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/NullTestCase.xtend
npasserini: I think that the message should be "receiver of || is null", emphasizing
that the error is the null receiver and not the message sent. Sorry :)

fdodino:
AAAAAAAGH! Crashed definition of == in Object!
Ok, I'll see it tomorrow.

fdodino:
Typo fixed.
Now, 

> do not know if you liked my previous proposal return other != null && self.equals(other)
Yes,  I liked.
I defined == as follows

``` java
		method ==(other) {
			return other != null && self === other
		}
```

Because original definition of == was

``` java
		method ==(other) {
			return self === other
		}
```

Because if you define 

``` java
		method ==(other) {
			return other != null && self.equals(other)
		}
```

There is a problem, since equals is a synonym for ==

``` java
method equals(other) = self == other
```

Thus,, if wollok object you want to use doesn't override equals this is a problem.
I tested this definition and it breaks : 21 errors and 399 failures.

npasserini:
My proposal is that equals stops being a synonym of == and that == takes
care of nothing/null values, so that user code does not have to handle it.

So implementations would be:
method ==(other) { return other !== nothing && this.equals(other) }
method equals(other) { return self === other }
method ===(other) native (?)

Then, users should
- never override == nor ===
- override equals if necessary.
- never send equals

flbulgarelli:
@npasserini that may work. 

We should explicitly tell that in the documentation.

npasserini:
@javierfernandes @tesonep @fdodino @matifreyre What do you think about this proposal?
Should I create an issue to change it?

matifreyre:
Nice!
Just a single question... should nothing == nothing succeed? Why?
(consider that I'm asking if it should, not if it does)

npasserini:
I think so, because the objective of nothing is exactly representing for
representing the absence of a value, so at one point we should allow to
check if something "is nothing".

At one point I was thinking about having a specific construct to check if
something is nothing, but I think that it complicates things even more...
so the best we can do is to allow for nothing supporting ==.

What do you thinl?

matifreyre:
:+1:

javierfernandes:
👍 
We may actually add a warning validation if you override == or ===

flbulgarelli:
> We may actually add a warning validation if you override == or ===

@javierfernandes should this be a warning harcoded in the language or could it be an annotation like java's `final`? 

> Should I create an issue to change it?

@npasserini see this issue https://github.com/uqbar-project/wollok/issues/705

flbulgarelli:
> At one point I was thinking about having a specific construct to check if
something is nothing, but I think that it complicates things even more...
so the best we can do is to allow for nothing supporting ==.

:+1: It doesn't have sense to add new control structures for `nothing` if we are planning to remove it eventually.

javierfernandes:
I don't like the "final" concept in languages.
I suffered many times the need to extend a class which had final (or
private) methods in Java and it was horrible.

I think that visibility not always involves the current class but it's like
a two-party decision between the class and whoever is
going to use it / extend it.
Same idea for checked exception.
Otherwise the class is like taking too much impact client code with its
decisions.
I prefer being open.

Although in another subject, I do think that it would be good to have a
visibility modifier for internal methods
that are not intented to be used directly from "outside" the object.
Kind of a "protected" in Java.
But the default should remain being like "public".

Probably just to be used by the SDK / framework.

Also I was wondering if modelling the "alias" as a language feature could
avoid this kind of confusion in inheritance
and override of methods.

Something like

```xtend
method equals(other) { .... }

alias method ==() for equals(other)

alias method ==() for equals   // if there's only one equals you can avoid
the arg
```

alias method

   - don't have a body
   - MUST refer to an existing method
   - cannot be overriden. You can only override the aliased method
   (original)


Any thoughts ?

Another syntax

```xtend
method equals || == (other) {
}
```

```xtend
method equals() with alias == {
}
```

```xtend
method equals(other) {
}
with alias ==
```

npasserini:
On Thu, Jul 7, 2016 at 4:37 PM, Franco Leonardo Bulgarelli <
notifications@github.com> wrote:

> At one point I was thinking about having a specific construct to check if
> something is nothing, but I think that it complicates things even more...
> so the best we can do is to allow for nothing supporting ==.
>
> 👍 It doesn't have sense to add new control structures for nothing if we
> are planning to remove it eventually.
>

I don't understand. Maybe I lost something, but I do not think that there
is any consensus on throwing "nothing" away, more like that the consensus
is to keep the nothing.

It needs more discussion, but I understand that the way we were agreeing on
is to separate:
1) nothing as an explicit value which is a specific object to represent the
absence of a value.
2) an uninitialized reference, which does not point to an object and it is
an error to try to send any message to it (including == or !=).

npasserini:
On Thu, Jul 7, 2016 at 5:00 PM, javierfernandes <notifications@github.com>
wrote:

> I don't like the "final" concept in languages.
> I suffered many times the need to extend a class which had final (or
> private) methods in Java and it was horrible.
>
> 100% agree on that, no final. Just a comment "this method is not intended
to be overridden" if you do it is at your own risk.


On the rest... I kind of dislike aliases, I would not have lots of them, so
adding specific features for aliases is not my favorite path to walk.

Moreover, I proposed to add specific semantics to == (i.e. handling
nothing) so if you agree to follow with that idea, == and equals would not
be aliases any more.

flbulgarelli:
> I don't understand. Maybe I lost something, but I do not think that there
is any consensus on throwing "nothing" away, more like that the consensus
is to keep the nothing.

OK, ok... replace _planning removing_ by _debating removing/complementing/changing semantics_ :wink: : In any case we should be conservative with such control structures as long as we don't finish that discussion, shouldn't we?

flbulgarelli:
Regarding the `final` discussion I am not proposing an strict mechanism for avoiding you override anything. But the problem is that comments are not processed by the tool, and you need to make custom checks. 

My proposed semantics of `final`- which I would call somenthing like `template` - thinking in the template pattern -  or something like `please do not override me unless you know what you are doing` :stuck_out_tongue:  is exactly that: just add display a warning that says "you probably don't want to do this".

flbulgarelli:
I actually would propose the same semantics for a `private` : it just suggest you to not send that message , but it doesn't forbid you to do that. 

Then tools like Eclipse can use that information

npasserini:
On Thu, Jul 7, 2016 at 5:30 PM, Franco Leonardo Bulgarelli <
notifications@github.com> wrote:

> I don't understand. Maybe I lost something, but I do not think that there
> is any consensus on throwing "nothing" away, more like that the consensus
> is to keep the nothing.
>
> OK, ok... replace *planning removing* by *debating
> removing/complementing/changing semantics* 😉 : In any case we should be
> conservative with such control structures as long as we don't finish that
> discussion, shouldn't we?
>
In any case, given that we decided that nothing is an object, I do not see
why we would have specific language constructs for it. So I think we can
agree in not having those constructs, and leave the jedi mental tricks
later.

Issue #668, opened 2016-06-01
Enable REPL while Wollok-Game runs


PalumboN:
Running a `.wlk` file with the REPL and execute `wgame.start()` it opens the Wollok-Game screen ok, but the console stays disabled. I think that it's waiting to WG threat finish.

PR #777, opened 2016-06-28
Dev #668 enable repl while wollok game runs

PalumboN:
At now, we must join the Wollok-Game thread to Wollok `mainLoopThread.join` because Wollok invoke exit call system `System.exit(0)` and we need block it.

But when Wollok-Game is running from REPL we want continue executing code, not block the REPL.

At first I thought that if wgame native implementation can know from where Wollok is running it can start the app like expected. But only WollokLauncher know this, no the interpreter.
So I thought we can configure the Gameboard if Wolllok is running from REPL:
```XTend
if (parameters.hasRepl) {
	// Configure Wollok-Game
	new WollokRepl(this, injector, interpreter, mainFile, parsed).startRepl
} 
```
But this implies WollokLauncher knows Wollok-Game module. I don't know if it is a good decision. Can you help me?

For now, I create other method for start the app: `wgame.run()` that it's non-blocker. So we can run Wollok-Game from de REPL with this message, and with `wgame.start()` from programs.


tesonep:
I understood that from the REPL you should send the #run and from your code the #start??
I think the API should be the same, or why not?

Why not adding a way of detecting how we are running (REPL, Tests, or normal)?
If you think that can be useful, we can have a Wollok object knowing this

What do you think?

PalumboN:
Yes! Having that information in the Wollok world it's ok for me. 
But I don't know how do this without dirtying the design.

tesonep:
I think having an object to interact with the native world can be helpful. Even from the wollok side, I do not known what @javierfernandes thinks about it.

tesonep:
I have added runtime object in the package wollok.vm
It's not visible by default, by now you can ask if you are running in interactive mode (with the REPL), I have added that to the Interpreter too, so we can have that information also in the Native side.

Please check the changes.

tesonep:
@PalumboN have you checked this? I have broke everything? What do you think I have changed in the code?

PalumboN:
@tesonep, Hi! Sorry for waiting, I was in travel. Thanks! This is just what I want. 

The code is ok for me. But I run it and fails. I don't have the error right now but I it's possible that wgame doesn't found runtime object, can be? It is working in your PC?

Anyway, at night I pull dev to this branch because the conflicts and try it again.

PalumboN:
I pull dev and now it's working. @tesonep check it, I have to create a new `.wlk` for runtime object and add the import in `lib`.
Can merge them, if all it's ok for you.

Issue #710, opened 2016-06-08
Exceptions thrown with self.error are different from those thrown by throw


flbulgarelli:
See the following code: 

```
program abc {
	throw new Exception("foo");
	console.println(null + "hello")
}	
```

It nicely produces: 

```
wollok.lang.Exception: foo
	at  [/program.wpgm:23]
```

But, the following: 

```wollok
program abc {
	self.error("foo")
	console.println(null + "hello")
}
```

Produces a strange error: 

```
org.uqbar.project.wollok.interpreter.WollokInterpreterException: Error evaluating line file:/home/franco/Downloads/wollok-linux.gtk.x86_64%20(1)/workspace/mi-proyecto/src/program.wpgm:[22]: program abc {
	self.error("foo")
	console.println(null + "hello")
}
	at org.uqbar.project.wollok.interpreter.WollokInterpreter.eval(WollokInterpreter.java:248)
	at org.uqbar.project.wollok.interpreter.WollokInterpreterEvaluator.eval(WollokInterpreterEvaluator.java:132)
	at org.uqbar.project.wollok.launch.WollokLauncherInterpreterEvaluator._evaluate(WollokLauncherInterpreterEvaluator.java:68)
	at org.uqbar.project.wollok.launch.WollokLauncherInterpreterEvaluator.evaluate(WollokLauncherInterpreterEvaluator.java:212)
	at org.uqbar.project.wollok.interpreter.WollokInterpreterEvaluator.evaluate(WollokInterpreterEvaluator.java:1)
	at org.uqbar.project.wollok.interpreter.WollokInterpreter.interpret(WollokInterpreter.java:149)
	at org.uqbar.project.wollok.interpreter.WollokInterpreter.interpret(WollokInterpreter.java:123)
	at org.uqbar.project.wollok.launch.WollokLauncher.doSomething(WollokLauncher.java:43)
	at org.uqbar.project.wollok.launch.WollokChecker.doMain(WollokChecker.java:83)
	at org.uqbar.project.wollok.launch.WollokLauncher.main(WollokLauncher.java:31)
Caused by: org.uqbar.project.wollok.interpreter.WollokInterpreterException: Error evaluating line file:/home/franco/Downloads/wollok-linux.gtk.x86_64%20(1)/workspace/mi-proyecto/src/program.wpgm:[23]: self.error("foo")
	at org.uqbar.project.wollok.interpreter.WollokInterpreter.eval(WollokInterpreter.java:248)
[...]

PR #780, opened 2016-06-28
Fix issue #710

fdodino:
Changing error message 
Cannot use this in a program's code !
for 
Cannot use reference to self in a program's code !


flbulgarelli:
@fdodino nice, but this this solve issue #710 ? It looks like a quite different problem. Or this this issue a side-effect of changing the messages :scream:

fdodino:
Sorry , I misunderstood the problem.

flbulgarelli:
Ok!


But it is still a fix :smile:

Issue #431, opened 2015-11-17
Split wollok SDK-lib library into several files


javierfernandes:
Currently we just have on wollok.wlk but it is starting to grow a lot.
We need to split it into different files.

javierfernandes:
This depends on refactoring / rethinking our imports system.

I wanted to split it like this:

```bash
wollok
├── lang.wlk
└── lib.wlk
```

But the classes/objects FQN is only taking into account the file name, and not the file path.
```java
Could NOT find wollok.lang.Object in scope: [lang.Exception, lang.StackTraceElement, lang.Object,
```
See that the scope is being created as "lang.Object", etc. instead of "wollok.lang.Object"

The problem is hard because we cannot just take the folder and prepend its name to the FQN. Where should we cut navigating up the filesystem ?

Again I think that we need a way to define a "wollok project". Maybe for now it could be just an empty file (?) like ".wollokprj"  Like a marker file.
Having it without content also makes it easier to work on wollok without and IDE.

I think this is also the cause of the current problem where a wollok file cannot import objects/classes outside of the file folder.
I'm not sure the import system should rely on relative paths (relative to each file).
I guess that it is better to have absolute FQN's based on the projects root.

@tesonep 
@npasserini 

?

PR #787, opened 2016-07-05
Fixing the implementation of Imports

tesonep:
This includes a number of things:

- Divides the Wollok SDK in different files (#431)
- Allows the use of different nesting packages
- Adds a cache in the lookup of manifest and root marker file.


fdodino:
Awesome!!

org.uqbar.project.wollok.lib/src/wollok/lib.wlk
```diff
@@ -0,0 +1,130 @@
...
+
+class StringPrinter {
+	var buffer = ""
+	method println(obj) {
+		buffer += obj.toString() + "\n"
+	}
+	method getBuffer() = buffer
+}	
+
+object wgame {
```
javierfernandes: Should we have wgame as the game file ?
I suspect you put it here to have it automatically imported.
Now, as wollokgame is an "advance" feature, maybe we won't mind having an import. Right ?
Maybe imports where already explained to this users when they get to write wollok game

org.uqbar.project.wollok.ui/src/org/uqbar/project/wollok/ui/wizard/WollokDslNewProject.xpt
```diff
@@ -28,5 +28,9 @@ log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=%5p [%t] (%F:%L) - %m%n
 «ENDFILE»
 
+«FILE "src/WOLLOK.ROOT"-»
```
javierfernandes: This is the new file that determines where the root code is for project-relative imports, right ?

In the future I think that it would be nice to model it in a file like a pom, including data as version and dependencies, etc.
Kind of a pom.xml or package.json.

org.uqbar.project.wollok/src/org/uqbar/project/wollok/scoping/cache/WollokGlobalScopeCache.xtend
```diff
@@ -0,0 +1,11 @@
+package org.uqbar.project.wollok.scoping.cache
+
+import java.util.List
+import org.eclipse.emf.common.util.URI
+import org.eclipse.xtext.xbase.lib.Functions.Function0
+import org.uqbar.project.wollok.wollokDsl.Import
+import org.eclipse.xtext.resource.IEObjectDescription
+
+interface WollokGlobalScopeCache {
```
javierfernandes: This is to fix the other issue reported by @Juancete about the imports files being loaded many times, right ?

javierfernandes: How's the caching strategy ?
Does it cache all imports ? or just the libraries outside of the project ?
I'm worried about content being cached and not being invalidated when the imported file changes.

org.uqbar.project.wollok.ui/src/org/uqbar/project/wollok/ui/wizard/WollokDslNewProject.xpt
tesonep: Yes it is the root file. 
Maybe it is a good idea to do the change now. 
Now thinking there is a file called wollok.manifest with the exported .wlk files. I will
unify both.

wollok-tests/src/wollok/imports/packageB/file2.wlk
```diff
@@ -0,0 +1,5 @@
+import packageA.file1.x
+
+object y{
+
+}
```
javierfernandes: Is there any test for project-relative imports like


project
  src
      wollok.root
      afolder
            afile
      bfolder
            bfile

and afile imports

import bfolder.bfile 

?

org.uqbar.project.wollok.lib/src/wollok/lib.wlk
tesonep: I don't know, I haven changed the distribution in packages because everything will destroy. I don't know what do you think @Juancete @PalumboN and the others?

tesonep:
I will divide the collections also, because as they are automatically imported nothing is broken, and also unify with the wollok.manifest

org.uqbar.project.wollok/src/org/uqbar/project/wollok/scoping/cache/WollokGlobalScopeCache.xtend
tesonep: There is two caches:

- One for the location of the "root file", the key here is the string representation of the path of the containing directory, so if there are many files in the same folder it will resolve only once. If you move the .wlk file to other directory it will recalculate for that directory. If you move the "root file" everything is doom. That is why I have put it hidden. I rather prefer not to be visible. 

- The other cache is for the imported objects, this cache use as key the string representation of the URI. In the cache I'm storing an object having the URI, the imports and the EObjectDescriptions. So before returning the value the list of imports is checked. If the list of imports changes the imported references are reloaded.

Now I am thinking if you change the imported file in the other world, it maybe is not updated. Maybe I should invalidate all the cache when a file is saved. I will do that.

wollok-tests/src/wollok/imports/packageB/file2.wlk
tesonep: I don't know if relative imports are working, maybe I should try them, I have not intended to do that.

tesonep:
After reading the comments of Javi, I have realise a problem with the cache, so I will fix it. And also I will not unify with the wollok.manifest, because the root file should not be moved or edited, or even should be visible for the user.

org.uqbar.project.wollok.ui/src/org/uqbar/project/wollok/ui/wizard/WollokDslNewProject.xpt
javierfernandes: 👍

org.uqbar.project.wollok/src/org/uqbar/project/wollok/scoping/cache/WollokGlobalScopeCache.xtend
javierfernandes: Yes. I don't  know of a way besides having your own eclipse "Builder" or BuildParticipant to be notified about a file changed / being built.

If I remember correctly there is an xtext specific extension point for extending this. Something about BuildParticipants

wollok-tests/src/wollok/imports/packageB/file2.wlk
javierfernandes: Oh, ok, so what are the new cases that this import mechanism support ?

javierfernandes:
But if you add a build listener to invalidate the cache, you can also use that for detecting a change in the wollok.root. Right ?
So it won't be a problem if someone moves it or changes it.

tesonep:
The problem is that I don't want to be visible the file

tesonep:
I have added as a save action, we have already one (the auto format on save).

I have added the ability to have imports depending on the paths, but always absolute imports, not relatives (relatives can easily added later), the idea is that if you have:

- WOLLOK.ROOT (this is hidden)
- a 
  - file1.wlk
    - object x 
    - object y
  - file2.wlk
  - a.wlk
    - object q
- b
    - file3.wlk
      - object z

From file2 you can import:

import a.file1.x
import b.file3.z
import a.q

As you see from the package file2, you have to use the full package to import the siblings.

tesonep:
I have fixed the problem with the cache.

javierfernandes:
Nice. 
So, to summarize, lets see if I got it right.
If I do have a wollok.root file then all imports on the project are treated as absolutes (or well, relative to that file).
If I don't then it will have the old "relative to the current file" behavior.

Right ?

tesonep:
Yes, the wollok root file is created by the wizard in the new projects, for the old projects there is the compatibility behavior (it works as until today).

org.uqbar.project.wollok.lib/src/wollok/lib.wlk
npasserini: I agree on moving it, but it is not necessary to do it in the same refactor.

npasserini:
:+1: Merge

Issue #786, opened 2016-07-01
Copy & Paste from REPL console adds font special characters


fdodino:
If you copy messages from REPL console and paste it no matter what kind of text editor, special characters appear. It is annoying.

fdodino:
Example:

[36;1mWollok interactive console (type "quit" to quit):[m
[36m>>> [maisugrgyset

PR #788, opened 2016-07-05
Dev #786 repl copy paste

fdodino:
Fix issue #786 

Now you can copy from REPL console and paste it in a Plain Text Format.
Pending: RTF conversion, because SWT has a RTFWriter as an inner private class we can't use.
Several hacks were done, it was very difficult to achieve this.
Not proud, but... it works.
Maybe a solution can be an epic refactor of REPL console highlighter, but it exceeds my knowledge (i'll wait a good time for that)


org.uqbar.project.wollok.ui.launch/src/org/uqbar/project/wollok/ui/console/editor/WollokReplStyledText.xtend
```diff
@@ -0,0 +1,109 @@
...
+		} catch (SWTError error) {
+			if (error.code != DND.ERROR_CANNOT_SET_CLIPBOARD) {
+				throw error
+			}
+		}
+		return false
+	}
+	
+	def void setClipboardContent(int start, int length, int clipboardType) throws SWTError {
+		val boolean isGtk = getFieldValue("IS_GTK") as Boolean
```
tesonep: Hi Fer, are you testing this on windows? or just in linux? Because my only fear is that this only works with gtk.

tesonep:
It's a good enougth solution, I don't know if there is a better way. For me it's ok and works in linux and mac, let's try it on windows. But I will integrate it.

org.uqbar.project.wollok.ui.launch/src/org/uqbar/project/wollok/ui/console/editor/WollokReplStyledText.xtend
fdodino: Didn't try it on Windows (don't have a Windows installation).

tesonep: I will test it, I got a VM

PR #793, opened 2016-07-08
Fixes #680  glitch in editor corrupts the view

### LINKED ISSUE #680: Code gets lost when you press enter!
I was typing the following: 

```wollok
/**
 * This is an example wollok hello world
 */
object ex {
	
	method sayHi(){
		
		console.println('Hello Wollok!')
	}
	
	method sayBaz() {
		
	}
}


class Foo {
	
}

class Bar inherits Foo {
	method zaraza() {
		
	}	
}
```

Then I positioned the cursor after the  `}` of `sayBaz`. I loose the method. 

```wollok
/**
 * This is an example wollok hello world
 */
object ex {
	
	method sayHi(){
		
		console.println('Hello Wollok!')
	}
	

}


class Foo {
	
}

class Bar inherits Foo {
	method zaraza() {
		
	}	
}
```

Then I selected the code. The method went back, but extra newlines where inserted between `sayHi` and `sayBaz`.

I was not always able to reproduce the issue. It seems that the code is actually there, but the editor is not rendering it propertly. 

I am using XUbuntu, and Wollok 1.4.3

javierfernandes:
This fixes the horrible bug #680

Seems like if you put a document listener into an xtext document and you perform a task that takes some time (like layout the class diagram) screws other parts of the editor.

What I've done is, probably what should have been done always, to run the listener code in a separated UIJob scheduled to be run in another moment.
So that it won't interfere with the editor behavior.

It now works fine.

My only explanation for other cases when people saw this happening in xtext own editor (and not just in wollok) is that they also have some listener that, sometimes takes longer and screws it up).

Fixes #680

### DIFF
--- a/org.uqbar.project.wollok.ui.diagrams/src/org/uqbar/project/wollok/ui/diagrams/classes/ClassDiagramView.xtend
+++ b/org.uqbar.project.wollok.ui.diagrams/src/org/uqbar/project/wollok/ui/diagrams/classes/ClassDiagramView.xtend
@@ -71,6 +71,14 @@ import org.uqbar.project.wollok.wollokDsl.WollokDslPackage
 
 import static extension org.uqbar.project.wollok.model.WMethodContainerExtensions.*
 import org.uqbar.project.wollok.ui.diagrams.classes.parts.InheritanceConnectionEditPart
+import org.eclipse.ui.PlatformUI
+import org.eclipse.core.runtime.Platform
+import org.eclipse.ui.internal.UISynchronizer
+import org.eclipse.core.runtime.jobs.Job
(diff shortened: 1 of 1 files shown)


npasserini:
Really you fixed that!!!!
This is great!

npasserini:
:+1: Jajjaaj, just 19 lines of code, half of them being imports.

javierfernandes:
Yes, and the lower change to the handler method shouldn't be included, but.. just got lazy to remove it before commiting :P 
Just formatting because I commented it to detect the issue.

fdodino:
Champion! Champion! Champion!

flbulgarelli:
Awesome. Merging!

PR #794, opened 2016-07-09
Dev remove eclipse local files

### LINKED ISSUE #57: Remove versioned resources after maven build system
.project (?) .classpath (?), .settings (?)

javierfernandes:
Fixes #57.

I think this can contribute to have more stability in the dev environment.
As we are using maven right now we don't need this files, and local eclipse changes them often.
So it confuses code reviews and stuff like that.


javierfernandes:
The build error is related with this

[[https://www.eclipse.org/forums/index.php/t/459920/]]

Seems like they need the .project file for the build. Which sucks.
I need to see what he means by customizing the StandaloneSetup.
I'm not sure where's the code that looks for the .project file.

fdodino:
I think we should close this PR, @npasserini @PalumboN what do you think?

npasserini:
Yeah, I think I already did this when deleting xsemantics.

PR #800, opened 2016-07-10
Fixes #799 static check for self usage in a program

### LINKED ISSUE #799: [validator] Cannot use "self" in a program
Using self in a program body currently fails with a runtime error telling you that you cannot use self in a program.
We need to catch this as a static check

javierfernandes:


javierfernandes:
I'm thinking that this may also works for self in a test

Issue #774, opened 2016-06-28
Validation: Non boolean operands on boolean expressions or if


javierfernandes:
This are typically typing errors.
There are some grammar elements which require a boolean expression like OR and AND are binary operators which eventually need that both of their operands (left and right side) be a boolean or boolean producing expressions.

Without a type system and type inference we cannot statically check this, but we can check some basic cases like the followings:
(this applies to the condition part of the if, but also to and, or and not operations)

```xtend
/* XPECT_SETUP org.uqbar.project.wollok.tests.xpect.WollokXPectTest END_SETUP */

object pepita {
}

object p {
	method run() {
		// XPECT errors --> "Non boolean expression" at "null"
		if (null)
			throw new Exception("asd")
		
		// XPECT errors --> "Non boolean expression" at "2"
		if (2)
			throw new Exception("asd")
		
		// XPECT errors --> "Non boolean expression" at "2+2"
		if (2+2)
			throw new Exception("asd")
		
		// XPECT errors --> "Non boolean expression" at "2"	
		a = 2 && p != null
		
		// XPECT errors --> "Non boolean expression" at "2"	
		a = 2 || p != null
		
		// XPECT errors --> "Non boolean expression" at "p"	
		a = pepita || p != null
		
		// XPECT errors --> "Non boolean expression" at "new List()"	
		a = new List() || p != null
		
		
		// more cases with method call require a type system with inference
	}
}
```

javierfernandes:
Another one that is already validated in runtime by 61aefa8

```scala
var expresionNula
if (expresionNula)
```

npasserini:
There are some issues that we might allow as warnings. For example I
sometimes modify my code from

if (some condition) { blah }

temporarily to

if (true || some condition) { blah }
just for debugging purposes.. what do you think?

javierfernandes:
Yes could be a warning.

matifreyre:
Will it be configurable?
I'd like those as errors for students by default, but also that we can
change them for warnings for these cases.

npasserini:
Every validation should be configurable but in a global base, I do not know
if that is what you mean.

matifreyre:
Yes, that's it

javierfernandes:
Yes. They are global or per project.

What we dont have yet is an export import of theconf to a file or something

javierfernandes:
I've splitted this issue into two, this one and #803 

This one is for non-boolean values both in if condition and in boolean expressions.

[...]

PR #804, opened 2016-07-11
Dev #774 check non booleans in boolean exp

javierfernandes:
Checkout the xpect tests to understand the different checks.

### DIFF
--- a/org.uqbar.project.wollok/src/org/uqbar/project/wollok/validation/WollokDslValidator.xtend
+++ b/org.uqbar.project.wollok/src/org/uqbar/project/wollok/validation/WollokDslValidator.xtend
@@ -392,6 +392,15 @@ class WollokDslValidator extends AbstractConfigurableDslValidator {
 		if (variable.uses.empty)
 			warning(WollokDslValidator_VARIABLE_NEVER_USED, it, WVARIABLE_DECLARATION__VARIABLE, WARNING_UNUSED_VARIABLE)
 	}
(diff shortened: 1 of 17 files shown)


org.uqbar.project.wollok/src/org/uqbar/project/wollok/validation/WollokDslValidator.xtend
```diff
@@ -392,6 +394,86 @@ class WollokDslValidator extends AbstractConfigurableDslValidator {
...
+				report(WollokDslValidator_UNREACHABLE_CODE, it, WBINARY_OPERATION__RIGHT_OPERAND)
+			else if (leftOperand.isFalseLiteral)
+				report(WollokDslValidator_UNNECESSARY_CONDITION, it, WBINARY_OPERATION__LEFT_OPERAND)
+			// right
+			if (rightOperand.isTrueLiteral)
+				report(WollokDslValidator_ALWAYS_EVALUATES_TO_TRUE, it)
+			else if (rightOperand.isFalseLiteral)
+				report(WollokDslValidator_UNNECESSARY_CONDITION, it, WBINARY_OPERATION__RIGHT_OPERAND)
+		}
+	}
```
javierfernandes: This code is not beautiful, but we don't want to evaluate things twice. The validator is starting to grow, and we need to start thinking about performance.
Although if anyone knows a better way to write it. Be my guess

Issue #807, opened 2016-07-11
Incorrect stack trace frames order when an error is thrown within a block


flbulgarelli:
Normally, when an exception is thrown inside a block, you expect to see that failure at the top of the stacktrace. E.g:

```smalltalk
>  {2. 4} do: [ :x | self error: 'foo'  ]
Object: nil error: foo
Error(Exception)>>signal (ExcHandling.st:254)
Error(Exception)>>signal: (ExcHandling.st:264)
UndefinedObject(Object)>>error: (SysExcept.st:1415)
optimized [] in UndefinedObject>>executeStatements (a String:1)
Array(SequenceableCollection)>>do: (SeqCollect.st:827)
UndefinedObject>>executeStatements (a String:1)
```

Or: 

```ruby
irb(main):001:0> [1].each { raise 'foo'  }
RuntimeError: foo
	from (irb):1:in `block in irb_binding'
	from (irb):1:in `each'
	from (irb):1
	from /home/franco/.rbenv/versions/2.0.0-p481/bin/irb:12:in `<main>'
```

However, see what happens when you reproduce this situation in wollok: 

```
program tests { 
 [1].forEach { throw new Exception("foo") }
}
```
```
wollok.lang.Exception: Error while calling native method 'WCollection.fold(WollokObject, WollokObject) with parameters: null, aClosure(acc, e)
  at wollok.lang.List.fold(initialValue,closure) [/wollok.wlk:485]
  at wollok.lang.Collection.forEach(closure) [/wollok.wlk:264]
  at [__synthetic0.wpgm]
  used by: wollok.lang.Exception: Error while calling native method 'Closure.doApply(WollokObject[]) with parameters: [Lorg.uqbar.project.wollok.interpreter.core.WollokObject;@2867adb7
  at wollok.lang.Closure.apply(parameters) [/wollok.wlk:722]
  at wollok.lang.Collection.forEach(closure) [/wollok.wlk:264]
  at wollok.lang.List.fold(initialValue,closure) [/wollok.wlk:485]
  at wollok.lang.Collection.forEach(closure) [/wollok.wlk:264]
  at [__synthetic0.wpgm]
  used by: wollok.lang.Exception: foo
  at [__synthetic0.wpgm]
  at wollok.lang.Closure.apply(parameters) [/wollok.wlk:722]
  at wollok.lang.Collection.forEach(closure) [/wollok.wlk:264]
[...]

PR #811, opened 2016-07-12
Dev #807 exception in closures

javierfernandes:


javierfernandes:
@fdodino check out my change. I think that it could be good as "more knowledge" on the interpreter.
WollokProgramExceptionWrapper are the way the interpreter has to model wollok exceptions as java exception and therefore be able to throw them to reuse java exception mechanism (bubbling, catching, etc).
So if while calling a native method it throws one of those, then we can just rethrow it.
If it is another java exception, then it is a weird case (buggy native code) so we need to convert it into a wollok exception.

My code was wrong, and when it was a wollok exception wrapped, I was not treating it 100% incorrectly but I was introducing a new wollok exception with the original as inner exception.
That introduced a complex stack trace. Mixing native details with wollok code.

It was somehow useful though when developing natives.. but well, not for the common cases :P

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/natives/ExceptionPropagationInNativeTestCase.xtend
```diff
@@ -0,0 +1,39 @@
...
+	catch e {
+		assert.equals(
+"wollok.lang.Exception: SoundGarden
+	at  [natives.wpgm]
+	at natives.Repeater.ntimes(n,closure) [natives.wpgm]
+	at  [natives.wpgm]
+",e.getStackTraceAsString())
+	} 
+}
+		''']
```
npasserini: Why can't we take advantage of xtend ''' and use better indentation?

npasserini:
Looks good for me.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/natives/ExceptionPropagationInNativeTestCase.xtend
javierfernandes: Yeah the problem is that if I had tabs rich strings will take them as part
of the string.
I know that sometimes it recognize a piece of rich string is all indented
and ignore the first tabs, but it is tricky.
I don't fully understand how that works :S

I mean I can indent the first lines but not the ones that have the string
I'm comparing in the assertion, they will introduce spaces.

PR #818, opened 2016-07-13
Fixes #746 validate if a variable is only used to return it. Its a fi…

### LINKED ISSUE #746: Validation: Using local variables only for assignment and then return
For example this should be avoided:

```
method costos(persona, cant_hambiente, cant_pisos) {
	 var costo = empleados.sum({per=>persona.presupuesto(cant_hambiente,cant_pisos)})
	 return costo
}
method presupuestoMaximo(){
	var presupuesto_maximo = 0.20  * ahorros
	return  presupuesto_maximo
}
```

For this to be good it should have a metric to establish if the use is to improve readability. Maybe we can test if the right part of the assignation is more than sending only a message (It can be one or two, I don't know).

javierfernandes:
…rst approach. Limited let say

It also detected issues in other tests.
Wich I think make sense.


npasserini:
:+1:

Issue #819, opened 2016-07-13
Pasting text in REPL does not work


fdodino:
If you want to paste text previously copied in REPL console, it doesn't work.
You press ENTER after pasting, nothing happens.
You have to paste, then press space bar and then ENTER to make it work.
Multiple lines also doesn't work.

PR #820, opened 2016-07-13
Fixing issue #819 - pasting text in a REPL console

fdodino:
Tried with one and many lines. Now you can paste text previously copied in REPL console.
It's not platform dependant, fortunately.


fdodino:
Already tested on Windows (just in case).
Merging.

PR #821, opened 2016-07-14
Fix Issue #684

### LINKED ISSUE #684: Referenciable name must start with lowercase
When you type an object in uppercase, it is fine you get an error - and a quick fix, sweet! - but the error message is not clear. It should be something more user friendly, like _objects must be written in lower case_.

Juancete:
Close #684 a simple typo error


npasserini:
This proposal provides different messages for variables and parameters, why not have yet another option for constants?

PR #824, opened 2016-07-15
Multithreaded interpreter

flbulgarelli:


flbulgarelli:
Hi @npasserini, I was wondering if it is easy to allow clients to activate/deactivate wollok checks, and to change their severity. For example, something similar to this - format is not important, just I want to describe how I would like to be able to use it -: 

```json
{ "code": "program foo {}", 
  "type": "program|test",
  "locale": "es_AR", 
  ....other arguments...
  "checks": [ 
         {"name": "UNUSED_VARIABLE", "severity": "ERROR"}, 
         {"name": "OTHER_CHCK", "severity": "IGNORE"}
   ] 
}
```

Issue #817, opened 2016-07-13
commons.collection bundle not found


fdodino:
Update site from dev shows an error preventing downloads:

```
Your original request has been modified.
  "Wollok Language Feature" is already installed, so an update will be performed instead.
Cannot complete the install because one or more required items could not be found.
  Software being installed: Wollok Language Feature 1.5.0.201607120439 (org.uqbar.project.wollok.sdk.feature.group 1.5.0.201607120439)
  Missing requirement: Wollok Language Feature 1.5.0.201607120439 (org.uqbar.project.wollok.sdk.feature.group 1.5.0.201607120439) requires 'org.apache.commons.collections 3.2.2' but it could not be found
```

There is also an Eclipse issue for Wollok internal developers forcing to download org.apache.commons.collections manually.

fdodino:
@tesonep is working on, waiting for sourceforge...

fdodino:
Awesome! Many many thanks @tesonep. Can we document the steps to add a new bundle, to avoid having troubles next time?

PR #826, opened 2016-07-18
Fixing dependencies of commons collections

tesonep:
It fixes issue #817. 
Allowing the update of 1.4.x to 1.5


javierfernandes:
Was just missing in the feature.xml ?

tesonep:
Yes, basically the plugin should be inside the feature, if not is not dragged in the generation of the Update Site or in the installation. Basically Eclipse and Tycho uses the target platform to know from where to take the plugin. So It should be in both places, the feature.xml of the SDK and the target platform.

Issue #784, opened 2016-06-30
Prefer exceptions over special values on standard library


flbulgarelli:
Checking [this PR](https://github.com/uqbar-project/wollok/pull/782/files), I found that the following methods return special and weird values in scenarios that should fail:

* `String.indexOf` returns -1
* `String.lastIndexOf` returns -1

matifreyre:
The wrapped native String returns those values. We can just catch them in WString and throw an exception instead, right? Any voices against that?

PR #834, opened 2016-07-20
Dev #784 string indexes

matifreyre:
- Changes ```WString``` methods ```indexOf``` and ```lastIndexOf``` so that they throw exceptions when the element is not found. This also fixes both methods for elements to be found at index 0.
- Adapts method ```contains``` to the change on indexOf.


org.uqbar.project.wollok.lib/src/wollok/lang.wlk
```diff
@@ -1281,7 +1281,11 @@ class String {
 	 * 		"become".contains("CO")      ==> Answers false
 	 */
 	method contains(other) {
-		return self.indexOf(other) > 0
+		try {
+			return self.indexOf(other) >= 0
+		} catch e {
```
javierfernandes: I wonder if we should be catching the particular exception here. Otherwise it will be catching any error

javierfernandes:
For me this seems ready to merge.
G ahead

org.uqbar.project.wollok.lib/src/wollok/lang.wlk
flbulgarelli: :+1: We shouldn't  catch everything here. 

Also, the expression in the return can never return false, since now indexOf always returns   a number `>= 0`.

We could improve this code by either: 
   * preserving an internal version of `indexOf` with the original behaviour
   * by not using indexOf in the implementation of contains
   * by introducing a new, generic message like `{ self.indexOf(other)  }.throws(NoSuchElementException)`  which better reveals the intention of the code

flbulgarelli: I prefer the latter: 

```
method contains(other) = ! { self.indexOf(other) }.throws(NoSuchElementException)
```

matifreyre: :+1:  will do

matifreyre: Sorry... I forgot we still can't pass classes or classnames as parameters and resolve them in runtime. It's the same problem that we had with #666.

matifreyre: I'd like to find a clean way of doing it in pure Wollok, but I think we can switch to use a native contains from WString temporarily. The wrapped String already knows contains(...).

matifreyre: I did that change of switching to the native side.
If there are no complains, I'll merge it.

Issue #790, opened 2016-07-07
RTF clipboard copy support for REPL console


fdodino:
Previously... #786 

I fixed this (Windows installation pending), but still remains RTF clipboard copy support.
If you copy from REPL console and put it in Office Writer/Word, it only copies plain characters. That's because RTFWritter is inner private class of SWT.StyledText. There should be a RTFConverter between ANSI characters of REPL Console and RTF "tags".

fdodino:
Assignee should see here

WollokReplStyledText class

```java
	def void setClipboardContent(int start, int length, int clipboardType) throws SWTError {
		val boolean isGtk = getFieldValue("IS_GTK") as Boolean
		if (clipboardType == DND.SELECTION_CLIPBOARD && !isGtk) return
		val TextTransfer plainTextTransfer = TextTransfer.getInstance()
		//val plainTextWriter = createWriter("StyledText$TextWriter", start, escapedBlockText.length)
		//val String plainText = getEscapedDelimitedText(plainTextWriter)
		val plainText = content.getTextRange(start, length).deleteAnsiCharacters
		var Object[] data
		var Transfer[] types
		if (clipboardType == DND.SELECTION_CLIPBOARD) {
			data = #[ plainText ]
			types = #[ plainTextTransfer ]
		} else {
			// val RTFTransfer	rtfTransfer = RTFTransfer.getInstance()
			// rtf = buscar una libreria que convierta de ASCII a RTF
			//data = #[ rtfText,, plainText  ]
			//types = #[ rtfTransfer, plainTextTransfer ]
			data = #[  plainText  ]
			types = #[ plainTextTransfer ]
		}
		val clipboard = getFieldValue("clipboard") as Clipboard
		clipboard.setContents(data, types, clipboardType);
	}
```

PR #836, opened 2016-07-20
Repl #790 rtf copy

fdodino:
Finally, I got it.
Now you can copy in the REPL & paste in Word, Office Writer and any RTF document.


fdodino:
Oh, by the way, there are many CHANGELOG.md & changelog.txt, I suppose that is because I run a mvn clean install. They bother me, because it is difficult to see the changes, @tesonep @javierfernandes should I delete them? I thought there were part of the build, so I didn't remove them.

javierfernandes:
Right. Seems like maven should only generate this while releasing and also just one for the parent project, and not a file for each module :S

javierfernandes:
As Carlitos would say: "it is very difficult" to see the PR with all the changelogs. Github doesn't show them to me.
Can you see if you can delete the changelog files and commit ?

fdodino:
Ok, my Unix old days were back

```
# find * -print | grep changelog.txt | rm $1
```

just in case you need it.

fdodino:
travis

javierfernandes:
jajaja !!!

just in case an improvement to the command

find . -name changelog.txt | xargs rm -f


find . -name CHANGELOG.md | xargs rm -f

PR #839, opened 2016-07-24
Dev #666 assertion exception

### LINKED ISSUE #666: add expected Exception type to assert.throwsException

Juancete:
Fixes #666 

add the following methods:
```
// equals comparing exception
assert.throwsExceptionLike(new BusinessException("lalala"),{ => throw new BusinessException("lalala") } 

// check the message, no matter the type: (also useful for  initial i-dont-know-classes-yet testing )
assert.throwsExceptionWithMessage("lalala", { => throw new BusinessException("lalala") } )

// check the type, no matter the message:
assert.throwsExceptionWithType(new BusinessException(),{ => throw new BusinessException("lalala") } )
```

All methods that I created have a very similar algorithm. So to avoid repeating code, I added a method called throwsExceptionByComparing(block,comparison). The first parameter is the block to evaluate and throw an exception meanwhile  the second parameter is .... other block that receive a value (the exception) and compares with other value that you write in it's code.


PalumboN:
> Fixes #666
:smiling_imp:

PR #842, opened 2016-07-25
Adding validation for the library

### LINKED ISSUE #833: XTend static checks of Wollok code don't have the same validations as a running Wollok environment
I modified this method in lang.wlk due to previous changes to indexOf:

```xtend
	method contains(other) {
		try {
			return self.indexOf(other) >= 0
		} catch {
			return false
		}
	}
```

Notice how ```catch``` doesn't have a variable there. I didn't really need that variable, so I just tried it that way.  

There is a test for ```contains``` that goes like this:

```xtend
	@Test
	def void testContains() {
		'''
		assert.that("aguacate".contains("cat"))
		assert.that("aguacate".contains("agua"))
		assert.notThat("aguacate".contains("managua"))			
		'''.test
	}
```

This test passed perfectly, so I thought that that ```catch``` syntax was valid. ```test``` uses ```interpretPropagatingErrors``` and not ```interpretPropagatingErrorsWithoutStaticChecks```.

But when I tried to use that same syntax in the Wollok environment, I got a static check that didn't allow me to. See this simple code:

```xtend
	try {
		throw new Exception("text")
	} catch { 
		assert.that(true)
	}
	assert.fail("Error")
```

image

I didn't enable nor disable any check in the wollok environment, so I would expect to have the exact same checks. This case means there could be other tests that pass when they should fail.

tesonep:
I have added a test for the library. 
 - I have detected that we were not allowing the native methods to override other methods, I have removed that, because is working perfectly.
- I am not checking that the name of the Well Known objects starts with a lower-case letter because all the WollokGame constants starts with upper-case, maybe we should check this later.
- Fixes #833


PalumboN:
> I am not checking that the name of the Well Known objects starts with a lower-case letter because all the WollokGame constants starts with upper-case, maybe we should check this later.

:fearful: Maybe we should rename to lower case.

tesonep:
Today I will not change nothing, because that objects are in the documentation.

PR #897, opened 2016-08-03
Fixes #868 cross references in WKO instance variables initialization …

### LINKED ISSUE #868: Circular cross-references between WKOs fail in runtime
fails in run time, but editor dont warnings
is posible make it run?

```xtend
object juan {
	var mascota = firulais
}

object firulais {
	var duenio = juan
}
```

javierfernandes:
…produced a stack overflow

### DIFF
--- a/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/regression/RegressionTestCase.xtend
+++ b/org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/regression/RegressionTestCase.xtend
@@ -190,4 +190,20 @@ class RegressionTestCase extends AbstractWollokInterpreterTestCase {
 		'''.interpretPropagatingErrors
 	}
 	
+	@Test
+	def void bug_868() {
+		'''
+		object juan {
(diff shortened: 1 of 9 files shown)


javierfernandes:
Fixed in a branch without introducing the stack overflow control.
WKO's are now registered before as the first step and then initialized.
If an error occurr while initializing they are removed from the scope.

This is of course not thread safe. But enough for the moment.

Eventually it will be hard to make Wollok interpreter multithreading

@npasserini

javierfernandes:
@npasserini please review this because introduces multithreading unfriendly changes.
Just in case you can think of a better and still quick solution.
Otherwise just to be aware for further refactors of the interpreter

javierfernandes:
bump !

npasserini:
Ufff, sorry but for me it is reeeeally very difficult to pay a look at this
right now.
Maybe @tesonep can?

tesonep:
I have checked and the "global reference" is stored in the Interpreter, as a new interpreter is generated per each request, it should not affect the Wollok Server multithreading.

npasserini:
Thanks!

Issue #899, opened 2016-08-04
WKOs with incorrect image in Wollok Game are not printed


PalumboN:
When Wollok Game doesn't find a declared image the WKO isn't printed.

PalumboN:
Closed in dev-1.5 #915

Issue #898, opened 2016-08-04
Change default WKO image to new icon


PalumboN:
![screenshot from 2016-08-04 04 18 57](https://cloud.githubusercontent.com/assets/4098184/17394481/b854cdb6-59ff-11e6-92d7-e18e3f5cca0a.png)

PalumboN:
Closed in dev-1.5 #915

PalumboN:
@npasserini Are you talking about this old icon or the new icon in the PR: #915?

npasserini:
You're right, the new Icon looks great. Sorry.

PR #915, opened 2016-08-08
New default image for visuals

PalumboN:
Close https://github.com/uqbar-project/wollok/issues/898
Close https://github.com/uqbar-project/wollok/issues/899

When you don't declare an image:
default_image

And when you declare an image and it hasn't found:
not_found_image


javierfernandes:
Nice !
I'm afraid that using the wollok icon as it is makes it difficult to read. Since it has a transparency in the W.

Can you check this one ?
I'm not completely satisfied with it. But I was not able to make it more readable. Maybe if we use the other icon.

wgame-image-not-found

PalumboN:
Cool! Yeah I thought the same. But the text is added dynamically (it's not in the image), so I can use the same icon for both cases.

Do you have only the icon with white background?

javierfernandes:
wgame-image-not-found

Here !

PalumboN:
Done!
screenshot from 2016-08-10 23 03 30

What do you think now, @javierfernandes?

javierfernandes:
Cool ! sorry for the delay. I missed your last comment :S

PR #918, opened 2016-08-10
Map returns list

### LINKED ISSUE #916: Using map over a set should return a list
It's not useful to get a set as a result of a map, it's more logical to get a list to ensure duplicates are included. To me this is invalid (the collection returned by juegos() is a set of 2 objects):

> > > biblioteca.juegos().map({juego => juego.tiempoRestante()})
> > > # {30.0, 50}
> > > 
> > > timbaElLeon.jugar(20)
> > > biblioteca.juegos().map({juego => juego.tiempoRestante()})
> > > # {30}

PalumboN:
Close #916

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok/lang.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok/lang.wlk
@@ -427,11 +427,11 @@ class Collection {
 	 * Answers a new collection that contains the result of transforming each of self collection's elements
 	 * using a given closure.
 	 * The condition is a closure argument that takes a single element and Answers an object.
-	 * @returns another collection (same type as self one)
+	 * @returns another list
 	 * Example:
 	 *      const ages = users.map({ user => user.age() })
 	 */
-	method map(closure) = self.fold(self.newInstance(), { acc, e =>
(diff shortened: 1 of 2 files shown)


javierfernandes:
I understand this can be confusing. Not sure, probably the default in many languages is to return a list.
But, I also found useful to have a map method returning the same instance.
So for example if from a set of "Person"s I would like to get the ages, there's a change that I just want the ages without duplicates. For example to understand the "different" ages of this people.

Maybe having two methods ?

npasserini:
For me it is the same, as long as the documentation states it clearly
(which current proposal does).
- Either map returns always a list
- Or returns an instance of the same collection

Personally I find the first one more useful, but after participating in
several discussions about this topic, I understand that this is not the
most popular choice.
So I am willing to accept either of both.

I don't think that it is necessary to have both options, it is easy to get
the other behavior very easily in both cases:
a) If it returns always a list and you want a set:

mySet.map(...).asSet()

b) If it allways returns the same type but you want to keep duplicates:

mySet.asList().map(...)


I think that the first case is less probable, and that is the reason to
prefer to always return a list. But it is very difficult to have stats
about it, so it is just my gut... so do whatever you want.

PR #920, opened 2016-08-11
QuickFix create method in another file not working

### LINKED ISSUE #590: Quick fix to create a method doesn't work between different files
I have my test cases in a wtest and when I get this error "Method does not exist in well-known object" because an object doesn't define the message I'm sending, using the quick fix to create it will not add the template in the correct place.

In the test file, the template just appears in a random place above. I checked this from a defferent wlk file after importing the other definitions I had and the quick fix doesn't work either (the original object definition was not changed), only this time the contents of the file I clicked on the quick fix from was not changed (though a * appeared as if it had been edited).

tesonep:
Fixes #590.
@mmatos please see if it is working.

Thanks.


javierfernandes:
Nice !

PR #971, opened 2016-09-05
Fix #960

### LINKED ISSUE #960: Refactor REPL Highlight to avoid colliding styles
When you copy REPL console in RTF mode, there's a heavy algorithm based on a bug generated by

* WollokCodeHighLightLineStyleListener

		// add custom highlights
		calculator.provideHighlightingFor(resource) [ offset, length, styleIds |
			val styleId = styleIds.get(0) // just get the first one ??
			val style = stylesProvider.getAttribute(styleId)
			
			val s = new StyleRange(event.lineOffset + offset - headerLength, length, style.foreground, style.background)
			s.data = styleId
			 
			if (s.start <= originalText.length && s.end <= originalText.length) {
				styles.merge(s)
			}
		]

* WollokAnsiColorLineStyleListener

This generates duplicates in styles.
We should refactor & fix this so there's only one style.

fdodino:
This fixes undetected problems in style ranges for REPL console, causing bad behavior in RTF copies.

### DIFF
--- a/org.uqbar.project.wollok.ui.launch/src/org/uqbar/project/wollok/ui/console/editor/WollokStyle.xtend
+++ b/org.uqbar.project.wollok.ui.launch/src/org/uqbar/project/wollok/ui/console/editor/WollokStyle.xtend
@@ -107,58 +107,9 @@ class WollokStyle {
 	}
 	
 	def getStylesAtLine(int lineNumber) {
-		adjustStylesRange(getStylesSelected(lineNumber))
+		getStylesSelected(lineNumber).sortBy [ start ]
 	}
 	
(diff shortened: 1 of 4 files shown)


fdodino:
Worked beyond this issue

* Added a radio button for objects (default) & classes
* Validations: no numeric identifiers, you can't define an identifier beginning with numbers
* Objects starts with lowercase, Classes with uppercase
* Default Extensions are used

tesonep:
Hi @fdodino this PR is still relevant?, I think yes, but by the comments I don't know.

fdodino:
Yes, it is important! 
My previous comment was that I just made a commit for another branch, wrongly referencing this issue.
But it has an important fix for styles rendered in REPL.

Issue #945, opened 2016-08-28
Project name allows spaces but generates error


JessSaavedra:
![wollokerror](https://cloud.githubusercontent.com/assets/7197057/18030746/94f60676-6c99-11e6-97ca-c242f257f037.png)

lspigariol:
neither in files names.
the rules are same of all identifier, but is not checked or warning.

fdodino:
Ok, if you define "obj ects.wlk" file you cannot run it in a REPL console:
ERROR [main] (WollokChecker.java:89) - Repl can only be used with .wlk files.

PR #974, opened 2016-09-05
Fix wizards

### LINKED ISSUE #831: Autogenerated code inside new object doesn't match file name
Creating a new file for Wollok object named "someFile.wlk" generates the following code:

```Java
object abc {

}
```

Perhaps changing abc to the filename could be a better guess.
Cheers!

fdodino:
Fixed #831 #945 
Also fixed Wollok logo path in New Project Wizard.


fdodino:
Worked beyond this issue

* Added a radio button for objects (default) & classes
* Validations: no numeric identifiers, you can't define an identifier beginning with numbers
* Objects starts with lowercase, Classes with uppercase
* If no extension is specified you should not create a file

fdodino:
Worked beyond this issue

* Added a radio button for objects (default) & classes
* Validations: no numeric identifiers, you can't define an identifier beginning with numbers
* Objects starts with lowercase, Classes with uppercase
* If no extension is specified you should not create a file

And Sorry, no relation with PR #971

PR #981, opened 2016-09-11
Type system improvements

javierfernandes:
- Fixed TS tests cases an implementation after refactor we did to remove basic objects (like strings).
- Adding a new extension point for TypeSystem implementation.
- Adding a preference page to configure which type system impl you want (global / project)
- Making TS only show types on outline if enabled (was a bug)
- Making all TS tests parametric to run the same tests with any TS impl.
- Adding more TS tests and impl’s for xsemantics and substitutions
- Adding closure’s typing


javierfernandes:
The build fails but I'm not sure why :(
Maven log sucks.

I ran mvn test locally, and all xpect tests failed with NPE in xpect initialization.
Really weird

javierfernandes:
Ok, it is huge, hehe

I don't even remember what I did.
But I suspect that
* many changes were introduced just to update the branch with how wollok evolved on dev (for example syntax changes like inherits, etc)
* then I introduced a major feature some months ago to have the TypeSystem impl configurable through preferences.
* I also worked on having shared tests between the different TS impls.

Regarding your changes, not much to add. I understand that they improved the way to assert for errors by using the validator and the errors it reports, instead of our own exceptions / errors.
That is good because it allow us to have more "integral" tests.

What is important is that we need to merge this to be able to start working on smaller changes and avoid such a long divergence with dev.

npasserini:
So, do I merge it?
I have already another bunch of changes for another pull request.

javierfernandes:
yes

Issue #1006, opened 2016-10-02
Problems with target platform.


npasserini:
I installed the latest Eclipse Neon and I get problems with target platform.
After trying the instructions in [Wollok developers page](https://github.com/uqbar-project/wollok/wiki/Development---Environment-Setup), error did not disappear. 

There is a conflict because it tries to load two different versions of plugin "org.eclipse.pde.feature.group" (3.12.0 and 3.12.1). 

I couldn't find exact explanation for the origin of those requirements, but I infer that one comes from my eclipse version, selected location in target platform is 
`<repository location="http://download.eclipse.org/eclipse/updates/4.6/"/>`

Which matches my eclipse Neon instalation:
Eclipse DSL Tools
Version: Neon Release (4.6.0)
Build id: 20160613-1800

But I find also another one which might be the origin of the problem: `<repository location="http://download.eclipse.org/releases/neon/"/>`

javierfernandes:
Merged

PR #1007, opened 2016-10-02
Possible fix for #1006. Remove conflicting location for target platform.

npasserini:
After removing the conflicting dependency Eclipse complaints that we are using a target platform which is newer than my eclipse instalation. I do not how to fix that. Everything seems to work, though.


javierfernandes:
I've tested it with 

```
Version: Neon Release (4.6.0)
Build id: 20160613-1800
```

And works fine.
Indeed there were some "units" repeated both in the "neon" as well as the "updates" locations.

I'm not getting any error or warning about using a newer target platform. Even if my neon eclipse has some months now (I haven't updated it since a couple of months)

PR #1023, opened 2016-10-17
design by contract: invariants & pre/post-conditions

javierfernandes:
There are 2 main features on this PR
- Contracts: invariants, method requirements (preconditions), and postconditions (ensures). Applies to Classes, mixins, WKOs, and object literals.
- Configurable language: starting to work on a capability to have elements of the language enabled / disabled. The first implementation is purely based on grammar elements. So you can disable individual grammar rules (defined in .xtext file), and any usage. For a disabled grammar element it will: remove autocomplete, remove syntax highlight, implement an static check with error.

There are some other changes to fix tests and misc tasks that arised probably because of this branch took a couple of months. I fixed many "examples" tests that were failing (probably not running ?) after changes to the SDK


fdodino:
El PR quedó `stale` :(
Me parece que lo mejor es cerrarlo y dejar el branch para cuando lo podamos retomar.

Issue #1015, opened 2016-10-07
irksome error message in editor


lspigariol:
![image](https://cloud.githubusercontent.com/assets/11738011/19195334/90159708-8c87-11e6-9772-27718df91ea5.png)

i suspect this error is related to errors in projects (files moved, renamed, etc). 

now, ocurs me when save files.
in other ocassions, ocurs while typing, but cant reproduce it

fdodino:
Yes, it is irksome (hahah, I had to look its definition).

I've been struggling with this all weekend. It  is a classpath problem refreshing a wollk file (test, program) from wollok.ui.launch instead of wollok.launch, which can find wollok lib files. I still don't know if we can solve it by
- using "platform:/resource" to load wollok file from refactor, quick fix, etc.
- or adding wollok libs files (lang.wlk, lib.wlk) from wollok.ui.launch.
Maybe neither of them
:smile:

fdodino:
Related to #912

fdodino:
Ok, problem is here

ClassloaderClasspathUriResolver.findResourceOnClasspath - line 59 for resources beginning with "classpath:". 

Instead of using
org.eclipse.xtext.ui.util.JdtClasspathUriResolver
that resolves perfectly in XTextResourceSet:
URI result = XtextResourceSet.this.resolveClasspathURI(uri);

When you open a lib.wlk or lang.wlk, it uses ClassloaderClasspathUriResolver that leads to that error.
I was working on another branch, but in order to show hyperlinks to files I came to this.
For now I put a cache, so we're not loading n times wlk classpath files, and it opens very fast a lib.wlk
But if you close IDE and start it again, error appears again.

fdodino:
Ah, ah, ah...

```java
    public IClasspathUriResolver getClasspathUriResolver() {
        if (resolver == null) {
            resolver = new ClassloaderClasspathUriResolver();
        }
        return resolver;
    }
```

here, in XTextResourceSet.class, resolver is null (is not injected).

fdodino:
Well, this moment could be named "Breaking bad".
This morning I got a working solution.
I was so happy.
But... a few hours later nothing worked. Eclipse failed. Unit tests failed. And Wollok Tests in IDE didn't launch at all. Several classpath errors appeared in Wollok Console (instead of Developer Console).

I'll describe my solution and I guess why did it happen, just in case @javierfernandes or @tesonep have any better ideas.
[...]

Issue #912, opened 2016-08-05
A lot of errors while opening an SDK file !


javierfernandes:
If I try to navigate to  an object that is built in in the SDK (lang.wlk, etc) then it takes a lot of time. Still it opens it but on the first eclipse I can see a lot of this errors

``` java
Caused by: java.lang.RuntimeException: Error while loading resource [classpath:/wollok/lang.wlk]
    at org.uqbar.project.wollok.scoping.WollokGlobalScopeProvider.loadResource(WollokGlobalScopeProvider.java:210)
    at org.uqbar.project.wollok.scoping.WollokGlobalScopeProvider.lambda$6(WollokGlobalScopeProvider.java:185)
    at org.uqbar.project.wollok.scoping.WollokGlobalScopeProvider$$Lambda$16/1426317832.apply(Unknown Source)
    at org.eclipse.xtext.xbase.lib.internal.FunctionDelegate.apply(FunctionDelegate.java:42)
    at com.google.common.collect.Lists$TransformingRandomAccessList.get(Lists.java:572)
    at java.util.AbstractList$Itr.next(AbstractList.java:358)
    at com.google.common.collect.TransformedIterator.next(TransformedIterator.java:48)
    at com.google.common.collect.Iterators$5.hasNext(Iterators.java:543)
    at com.google.common.collect.Iterators$5.hasNext(Iterators.java:542)
    at com.google.common.collect.Iterators$5.hasNext(Iterators.java:542)
    at com.google.common.collect.Iterators.addAll(Iterators.java:356)
    at com.google.common.collect.Lists.newArrayList(Lists.java:146)
    at com.google.common.collect.Iterables.toCollection(Iterables.java:334)
    at com.google.common.collect.Iterables.toArray(Iterables.java:312)
    at org.eclipse.xtext.xbase.lib.Conversions.unwrapArray(Conversions.java:96)
    at org.uqbar.project.wollok.scoping.cache.MapBasedCacheContent.<init>(MapBasedCacheContent.java:36)
    at org.uqbar.project.wollok.scoping.cache.MapBasedWollokGlobalScopeCache.get(MapBasedWollokGlobalScopeCache.java:28)
    at org.uqbar.project.wollok.scoping.WollokGlobalScopeProvider.importedObjects(WollokGlobalScopeProvider.java:75)
    at org.uqbar.project.wollok.scoping.WollokGlobalScopeProvider.getScope(WollokGlobalScopeProvider.java:55)
[...]

PR #1024, opened 2016-10-18
First solution: cache of lib wollok objects

fdodino:
Although it is a work in progress, I prefer to open a PR to debate  certain implementations and enhacenments.

See #1015 | #912 

For @tesonep  and @javierfernandes 

* I'd like to obtain currentProject from editor (or any launcher) to successfully get classpath resources, like classpath:/wollok/lang.wlk. For now it works if this .wlk is imported from a "real" wlk or wtest file, developed by a student/teacher. If you open a lib.wlk without a wtest or wpgm or wlk developed by a user, it can't obtain the resource.
* Maybe I should replace my implementation of WollokResourceCache in favor of MapBasedWollokGlobalScopeCache.  I'll give a try.


fdodino:
I'm deleting this branch because I'm integrating it with branch wtests-rmi-buffered

PR #1025, opened 2016-10-19
Fix #904 - Decimal numbers operations

### LINKED ISSUE #904: wollok.lang.Exception: For input string: "0E-8".
Try the following: 

``` wollok
>>> 6.2500000 * 0.5 * 0
wollok.lang.Exception: For input string: "0E-8"
    at wollok.lang.Double.*(other) [/lang.wlk:1141]
    at  [/__synthetic0.wpgm:3]
    at  [/__synthetic0.wpgm:3]
    at  [/__synthetic0.wpgm:3]
    at  [/__synthetic0.wpgm:3]
    at  [/pepita.wlk:1]
```

Clearly the student did not explicitly introduce that expression, but it was the result of sending some message in a exercise I did. 

It looks like a - masked - precision error regarding the fact that we are using big decimals, right?

fdodino:
Original issue #904 
- Using a common definition in WNumber for decimal operations: add. subtract, multiply, divide, remainder
- Created several test cases for decimal numbers.

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok/lang/WDouble.xtend
+++ b/org.uqbar.project.wollok.lib/src/wollok/lang/WDouble.xtend
@@ -20,21 +20,33 @@ class WDouble extends WNumber<BigDecimal> implements Comparable<WDouble> {
 
 	@NativeMessage("+")
 	def plus(WollokObject other) { operate(other) [ doPlus(it) ] }
-		def dispatch Number doPlus(Integer w) { wrapped + new BigDecimal(w) }
-		def dispatch Number doPlus(BigDecimal w) { wrapped + w }
+		def dispatch Number doPlus(Integer w) { 
+			this.add(wrapped, new BigDecimal(w))
+		}
+		def dispatch Number doPlus(BigDecimal w) {
+			this.add(wrapped, w) 
+		}
 		//TODO: here it should throw a 100% wollok exception class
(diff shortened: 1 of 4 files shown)


fdodino:
Well, now Wollok decimals are rounded HALF_UP to 5 decimals for operands and results of

* add
* subtract
* multiply
* divide
* remainder

fdodino:
Well, last change was made to allow using big integers. Something like

```
>>> var a = 1000000000000000000
```

Previously, Wollok Interpreter was throwing an ugly exception: "For String 1+E15" or something like that.
Now I catch NumberFormatException (knowing there is no chance to interpret an alphabetic input) and instantiates a DECIMAL instead of INTEGER.

Added new tests.

npasserini:
Why not use always BigDecimals?

fdodino:
I confess I was tempted to do so.
But it is a big refactor, I'll leave for another occasion (or for someone else).

npasserini:
Looks good to me. 

Just one question, where are all numbers being wrapped again as wollok numbers? 
(I couldn't find the definition of operate method. Is it there?)

fdodino:
Hi @npasserini , it is in WNumber

```xtend
	def operate(WollokObject other, (Number)=>Number block) {
		val n = other.nativeNumber
		if (n == null)
			throw new WollokRuntimeException("Operation doesn't support parameter " + other)
		val result = block.apply(n.wrapped)
		newInstance(result)
	}
```

Issue #1018, opened 2016-10-13
Wrong comparisons on sets


clombardi:
# {3,4}.equals(#{3}) returns true, you can try it in a REPL.

The assert.equals exhibits the analogous problem, e.g. 
  assert.equals(#{3,4}, #{3})
does not make a test red.

Please fix it ASAP, we have problems with student tests at UNQ.

PR #1026, opened 2016-10-19
Fix set comparison

fdodino:
See #1018.

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok/lang/WSet.xtend
+++ b/org.uqbar.project.wollok.lib/src/wollok/lang/WSet.xtend
@@ -23,8 +23,9 @@ class WSet extends WCollection<Set<WollokObject>> {
 	}
 	
 	override protected def verifyWollokElementsContained(Collection set, Collection set2) {
-		set.forall [ elem |
-			wrapped.exists[ it.wollokEquals(elem) ]
+		set2.forall [ elem |
+			set.exists[ it.wollokEquals(elem) ]
 		]
 	}
+
 }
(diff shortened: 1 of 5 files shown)


org.uqbar.project.wollok.lib/src/wollok/lang/WSet.xtend
```diff
@@ -23,8 +23,9 @@ class WSet extends WCollection<Set<WollokObject>> {
 	}
 	
 	override protected def verifyWollokElementsContained(Collection set, Collection set2) {
-		set.forall [ elem |
-			wrapped.exists[ it.wollokEquals(elem) ]
+		set2.forall [ elem |
```
npasserini: This implementation is not good enough, we should check for the size of the sets.

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/interpreter/ListTestCase.xtend
```diff
@@ -119,6 +119,9 @@ class ListTestCase extends CollectionTestCase {
 		'''
 		assert.notEquals(['Hello'], ['Hellou'])
 		assert.notEquals([4, 5, 9], [5, 4, 9])
+		assert.notEquals([4, 5, 9], [5, 4])
+		assert.notThat([4, 5, 9].equals([5, 4]))
+		assert.notThat([4, 5, 9].equals([5, 4, 9]))
```
npasserini: Add asserts the other way around:

assert.notThat([5, 4].equals([4, 5, 9]))

PR #1031, opened 2016-10-23
Dev 1.5 #1027 duplicated class name

### LINKED ISSUE #1027: Doesn't check a repeated class name if you don't define a package
If you do somethig like 

**>>>aFile.wlk**

```
class Bleh {
...
}
class Bleh {
...
}
```

Doesn't check a repeated class name. But if you do 

**>>>aFile.wlk** 

```
package aPackage {
   class Bleh {
   ...
   }
   class Bleh {
   ...
   }
}
```

Shows an error.

Juancete:
Close #1027. 

Packages and files are not the same, so the checker for a duplicated class doesn't work if you don't create a package. 

I have created a new checker who see the package of a class. If it don't have, the checker will search for other classes.


javierfernandes:
I replied to this from a mobile device but I think it got lost somewhere.

I was saying that it could be possible to implement a single check methods covering both cases (classes within a wfile or wpackage, or even for future potential cases, like classes within classes / objects)

You need to work at "ecore" level, kind of using the generica "AST" API of the objects.

```xtend
val container = c.eContainer  // avoid the WFile cast
val classes = file.eContents.filter(WClass)      // use the "eContents" (?) method that gives you the list of children nodes
```

Juancete:
![Javi!]
(http://spa.fotolog.com/photo/10/63/63/las__morriseras/1213672899032_f.jpg)

javierfernandes:
I think that you can mix it all into a single assertion by working at ECore
level.
You check a WClass. No othee Wclass with the same name should be in the
eContainer.

It doesnt matter if the container is a wfile or package.

Something like this

wclass.eContainer.eContents.filter(WClass).filter(name ...)

Juancete:
Now it's in a single method called _classNameCannotBeDuplicatedInFileOrPackage_. Doesn't care if the eContainer is a WFile or a WPackage. It filters by WClass and the resultant collection is filtered by name and isn't the same WClass object. 

Would this be correct?

Only check the type of eContainer to return a different error message, but i can make a single error message.

javierfernandes:
That's fine ! Merging

Issue #1040, opened 2016-11-08
Fix New Project Wizard


fdodino:
In development mode background title of New Project Wizard is white, and appears ok:
![image](https://cloud.githubusercontent.com/assets/4549002/20118711/56d4cb28-a5e4-11e6-8042-12f55c53d07c.png)

But final product background title is transparent, or gray, so it appears like this:
![image](https://cloud.githubusercontent.com/assets/4549002/20118688/3844a192-a5e4-11e6-8796-48b5fb4e93e4.png)

It happens at least in Linux

fdodino:
I guessed it was GTK+ configuration in wollok.ini, adding these lines solved my problem:

```
--launcher.GTK_version
2
```

@javierfernandes Do you know how to add them programmatically while generating Linux product?

npasserini:
I would also change the title, why WollokDsl and not wollok?

2016-11-08 23:23 GMT+01:00 Fernando Dodino <notifications@github.com>:

javierfernandes:
Not sure, but it should be close to the "product" xml file.

There is a way from the "visual editor" to define custom "config.ini' files.
http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.pde.doc.user%2Fguide%2Ftools%2Feditors%2Fproduct_editor%2Fconfiguration.htm

But I'm not sure if config.ini is the same as eclipse.ini.
I think that it is not. And that "eclipse.ini" is what they call "launcher
params"

Here this guy sets some lancheroptions in the product xml

http://stackoverflow.com/questions/9357551/eclipse-rcp-product-custom-config-ini

 <launcherArgs>
      <programArgs>-nl it</programArgs>
      <vmArgsMac>-XstartOnFirstThread
-Dorg.eclipse.swt.internal.carbon.smallFonts</vmArgsMac>
   </launcherArgs>

Could it be a light ?

fdodino:
Nice! Thanks, I'll check this next week

fdodino:
You are a "great" (groso)
:smile: 

Added launcher args in updatesite project, 

![image](https://cloud.githubusercontent.com/assets/4549002/20282240/b29b2896-aa92-11e6-9ee9-5848ce71ec6f.png)

and it worked. 
This solves Ubuntu launching problem.

fdodino:
Changed "New WollokDSL project" to "New Wollok project" also

PR #1042, opened 2016-11-14
New wollok wizard

fdodino:
Forcing to use GTK Launcher for Linux Installation. Fixing #1040 issue.


fdodino:
It's has also a "usability" label because in Ubuntu it is a necessary configuration to make Wollok work.

Issue #896, opened 2016-08-03
Model StackOverFlow error and control it at wollok (interpreter) level


javierfernandes:
This arised from #868  but also from older cases.

There's no way we can avoid 100% this issue in any user code so if he does it we must fail in a "wollok way"
Currently the interpreter fails at java level and shows a java exception.

We need to model the wollok.lang.StackOverFlowException/Error and control execution in the interpreter raising this error and showing a good wollok stack trace.

javierfernandes:
Example today

```xtend

object juan {
	method blah() {
		self.blah()
	}
}


program a {
	juan.blah()
}
```

Error

```
	at org.uqbar.project.wollok.interpreter.WollokInterpreterEvaluator.evalAll(WollokInterpreterEvaluator.java:140)
	at org.uqbar.project.wollok.interpreter.WollokInterpreterEvaluator._evaluate(WollokInterpreterEvaluator.java:595)
	at org.uqbar.project.wollok.launch.WollokLauncherInterpreterEvaluator.evaluate(WollokLauncherInterpreterEvaluator.java:168)
	at org.uqbar.project.wollok.interpreter.WollokInterpreterEvaluator.evaluate(WollokInterpreterEvaluator.java:1)
	at org.uqbar.project.wollok.interpreter.WollokInterpreter.eval(WollokInterpreter.java:237)
	at org.uqbar.project.wollok.interpreter.AbstractWollokCallable.eval(AbstractWollokCallable.java:131)
	at org.uqbar.project.wollok.interpreter.AbstractWollokCallable.lambda$0(AbstractWollokCallable.java:87)
	at org.uqbar.project.wollok.interpreter.AbstractWollokCallable$$Lambda$48/1311315651.apply(Unknown Source)
	at org.uqbar.project.wollok.interpreter.WollokInterpreter.performOnStack(WollokInterpreter.java:197)
	at org.uqbar.project.wollok.interpreter.AbstractWollokCallable.call(AbstractWollokCallable.java:99)
	at org.uqbar.project.wollok.interpreter.core.WollokObject.call(WollokObject.java:119)
	at org.uqbar.project.wollok.interpreter.WollokInterpreterEvaluator._evaluate(WollokInterpreterEvaluator.java:738)
	at org.uqbar.project.wollok.launch.WollokLauncherInterpreterEvaluator.evaluate(WollokLauncherInterpreterEvaluator.java:208)
[...]

PR #1048, opened 2017-01-11
Creating the stack overflow error from java's one

tesonep:
Fixing Issue #896 

I have implemented the StackOverflowError of Wollok from the JAVA. 

The problem with the implementation with limits is that it depends in the platform it is running and the implementation of the interpreter.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ install:
   - cmd: SET PATH=C:\maven\apache-maven-3.2.5\bin;%JAVA_HOME%\bin;%PATH%
   - cmd: SET MAVEN_OPTS=-XX:MaxPermSize=2g -Xmx4g
   - cmd: SET JAVA_OPTS=-XX:MaxPermSize=2g -Xmx4g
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -13,7 +13,7 @@ install:
   - cmd: SET PATH=C:\maven\apache-maven-3.2.5\bin;%JAVA_HOME%\bin;%PATH%
   - cmd: SET MAVEN_OPTS=-XX:MaxPermSize=2g -Xmx4g
   - cmd: SET JAVA_OPTS=-XX:MaxPermSize=2g -Xmx4g
-  - cmd: SET M2_HOME=C:\maven\apache-maven-3.2.5
+#  - cmd: SET M2_HOME=C:\maven\apache-maven-3.2.5
```
npasserini: This looks weird, why did we have a M2_HOME pointing to a C:\ path?

tesonep: I have never answer, but it was there before, and I just comment it. I don't know why It was there, but I think when I setup the windows CI I copied from somewhere.

Issue #739, opened 2016-06-14
Running REPL without a file asks for an open project


fdodino:
It shouldn't ask for a specific project.
If you have no projects, you cannot invoke REPL console.

PalumboN:
> If you have no projects, you cannot invoke REPL console.

:cry: Why?

fdodino:
I've seen this is not easy to achieve, because here in LaunchReplWithoutFileHandler you create a Run Configuration mode in Eclipse. 

But Eclipse needs a Project to run anything.

So... a workaround could be: ok, I create an empty project so I can run a console, but... I think it would appear an "Empty project" on Project Explorer. This would be confusing for any student.

A second thought: create an empty project without Java Nature, only for running REPL console. Filter this kind of project in Project Explorer but allow here to appear:

```scala
	def configureConfiguration(ILaunchConfigurationWorkingCopy it) {
		val root =  ResourcesPlugin.getWorkspace().getRoot()
		val openProjects = root.projects.filter[ project | project.isOpen() && project.hasNature(JavaCore.NATURE_ID)]
		
		if(openProjects.empty) {
			val x = new RuntimeException(Messages.LaunchReplWithoutFileHandler_notHavingWollokProject)
			DebugUIPlugin.errorDialog(DebugUIPlugin.shell, Messages.LaunchReplWithoutFileHandler_notHavingWollokProject, Messages.LaunchReplWithoutFileHandler_notHavingWollokProject, x) //
			throw x
		}
```		

What do you think @tesonep @javierfernandes @npasserini ?

tesonep:
One possible way of generating is generating the classpath and run the REPL as a "External Tool", just as it is a external program. The problem here is generating the classpath with all the elements. I think it can be worked in the same way the dependencies in a project, but It have to be checked. 
What do you think @javierfernandes ?

npasserini:
If we generate an empty project couldn't we hide it from the default
workspace?

npasserini:
I would really like to have a nice solution to this problem. @tesonep do you think we could do something?

tesonep:
I have to retry to generate the running as an external tool. I have seen some tools that performs this, I will try to be "inspired" from them to see how an external tool can refer to the elements installed in the plugins.

tesonep:
I have created a workaround for this. 
Basically I am creating (or reusing) an empty project, this project is hidden in the view. 
[...]

PR #1056, opened 2017-03-14
Running the REPL alone - Adding a workaround creating a hidden project

tesonep:
It fixes the issue #739.

I have created a workaround for this.
Basically I am creating (or reusing) an empty project, this project is hidden in the view.
I have chosen this approach because to run as a external tool, the whole classpath for the bundle should be calculated (it need the launcher, the lib, and all the xtext / eclipse dependencies). This is calculated in the build to generate the scripts to run from the console but they are not persisted. Even accesing to this classpath, it should be transformed to get the JARs from the plugins directory.

Maybe a better implementation can recalculate this classpath and store them for usage.

Please @fdodino, check if this is good enough.


fdodino:
Cool!

I tested it as a user, didn't see any line of code. But it works fine!

image

fdodino:
Thanks @tesonep !

Issue #1062, opened 2017-03-27
Problems running Xpect tests


fdodino:
Problems are in this tests

* WollokContentAssistTest, test proposals, because of absence of [closeQuietly method](https://github.com/google/guava/issues/1118 )
* WollokXPectTest 

For now both tests are ignored

Related xpect issues:
* https://github.com/meysholdt/Xpect/issues/185

tesonep:
We have seen with @fdodino that the Xpect is using Guava, but it is not declaring correctly (it's not declaring at all) the version to use. So, it is using the latest one and the latest one has a deprecated method. Maybe it is a good time to upgrade to xtext 2.11, xsemantics 1.11 and xpect  0.2.0.

fdodino:
+1

tesonep:
Only we should take care of:

**_Deprecated Components_
Old Xtext Generator**
With Xtext 2.9 a new generator infrastructure was introduced. In favor of this new generator the old generator has been deprecated. Bugfixes will be done primarily for the new generator and most of them will not be backported, thus it is highly recommended to migrate your languages’ .mwe2 workflows to use the new org.eclipse.xtext.xtext.XtextGenerator component.

**Running the Language Generator from Maven**
The Maven POM module org.eclipse.xtext.xtext, which has been deprecated since 2015, is no longer published. If you use the old generator component, you should replace the org.eclipse.xtext.xtext dependency according to the dependencies of that module. The new generator component requires much fewer dependencies; a suitable configuration for it can be obtained by creating a new Maven project with the Xtext project wizard.

tesonep:
Migrating to Xpect 0.2.0 did not fixes the problem with Guava deprecated method. 
I have been able to migrate to the new version of xtext, but some test are failing. 
I will see if I can make it work, but the Guava have to be solved using a different Guava version

PR #1065, opened 2017-04-12
Dev #1062 migrate to Xtext 2.11

tesonep:


tesonep:
I still need to migrate some classes, that now should been written in Xtend and they are written as Java so there are some "empty" xtend classes.

Issue #1039, opened 2016-11-07
sortBy should not return the resulting list


julian-berbel:
this is to coincide with the idea that a method should either have a returning value or side-effects, but not both.

lspigariol:
Fixed

PR #1078, opened 2017-04-25
Fix-#1039-sortBy-should-not-return-the-list

lspigariol:
Primer prueba de modificar algo de wollok. Que alguien lo revise


fdodino:
Sounds good to me!
I reviewed code, ran all the tests (1225 were ok), restarted travis build.
So, what do you think @tesonep ?

tesonep:
It looks great to me, just I remove the commented code, it is not adding anything (and always we can go back to Git)

tesonep:
I will merge it as soon as the CI finishes

PR #1091, opened 2017-05-29
fix #1052 exponentiation precedence 

### LINKED ISSUE #1052: Exponentiation should have higher precedence than multiplicative operations.
If you evaluate 

```program p {
	console.println(4 * 3 ** 2)
}
```

you will get 144 (=12^2) instead of 36(=4*9) as it should.

juanbono:
In order to fix the precedence bug I added a new rule to the grammar:
```
WExponentiativeExpression returns WExpression:
	WUnaryOperation (=>({WBinaryOperation.leftOperand=current} feature=OpPower) rightOperand=WExponentiativeExpression)?;

OpPower:
	'**';
```

Now `a * b ** c` equals `a * (b**c)` as expected.


fdodino:
Nice! Could you please add a new test?

Like

```xtend
// it should be 24 and not 216
3 * 2 ** 3
```

Thanks!

fdodino:
Grossum!!

PR #1093, opened 2017-05-30
fix-#1089 wrong error message when a test is defined in a .wlk file

juanbono:
I've defined a constant **_wollokDslValidator_TESTS_IN_FILE_** with the content of the error message.
Also I noticed that a "CLASSES_IN_FILE check" is missing.
Should I fix it in this PR or create a new issue?


tesonep:
I don't know if we want a check to check that classes are defined in a .wlk file, maybe you want to create classes in a test file. @npasserini @fdodino what do you think?

fdodino:
I didn't try to define objects or classes in a test file, because I didn't know it could be possible.
This year I went deep into tests and WTest definition and I realized this door was open.
I think we can let .wtest files to have objects and classes definitions, and avoid that check favoring simplicity (if you want to create a quick test without having to create 2 files).

Issue #1087, opened 2017-05-15
Add library support


lgassman:
Library support is needed in order to integrate wollok with other technologies, such as Arena.

PR #1095, opened 2017-06-02
Dev #1087 libraries

lgassman:
A library is a jar file with a wollok manifest file at root level.
if the jar is named "libname.jar", the manifest file must be named "libname.wollokmf"
A wollok manifest is a text file with one file path per line.
You can add another eclipse project as library to, In this case the project name must be the same that the wollok manifest: project: libname, manifest name: libname.wollokmf. The manifest must be in the root of a source folder.
you can add libraries to a wollok project using library property page. If you run a wollok project from wdk, you must use -lib obtion separated by coma.
example: -lib foo/foo.jar,bar/bar.jar -wf wollok.wpgm
A test library project was created. This project is compilated using a jardescriptor and it is putted on wollokLib folder of tests project. This library is used on LoadLibraryTest


lgassman:
We have a problem with the classloader on the continuos integration...

PR #1101, opened 2017-06-08
Dev #437 autocomplete show where method comes from

### LINKED ISSUE #437: Autocomplete should show where the method comes from
For example in Java/xtend eclipse shows as text decoration in gray where the method comes from

screen shot 2015-11-18 at 09 48 42

Juancete:
Shows FQN when autocomplete is called from "self". Now you can see where a method comes from.
Closes #437


Juancete:
@tesonep I think it will never happen. This feature only works when you write "self.", it means that you are _inside_ a class/object,etc. If you are using polymorphics objects, they are referenced by a variable, so you are using those objets from _ouside_ them and the feature will not work.

tesonep:
Ok, I didn't get it all. My fault, I should have run the code to see well what it was doing.

Juancete:
@tesonep Don't worry, no problem :+1:

Issue #1146, opened 2017-07-16
Type System: Integer  and Number resolution


fdodino:
Maybe this error will dissappear after #1129

This program

```xtend
object pepita {
	var energia = 0
	
	method volar(kilometros) {
		energia = energia + kilometros * 20	
	}
}
```

Leads to 
`expected <<Integer>> but found <<Double>>`
type system error.

npasserini:
@fdodino could you check if this is not already solved?

fdodino:
I'm sure this has already been solved. I don't know if we should add a test for this case.

PalumboN:
We have this test:
```xtend
class Golondrina { 
	// XPECT type at energia --> Number
	var energia = 100
	
	// XPECT methodType at vola --> () => Void
	method vola() { energia = energia - 10 }
	
	// XPECT methodType at come --> (Number) => Void
	method come(grms) { energia = energia + grms * 0.10 }
}
```

fdodino:
I'm back for Wollok. 

Ok, I changed the passing test to this

```scala
class Golondrina { 
	// XPECT type at energia --> Number
	var energia = 100
	
	// XPECT methodType at vola --> (Number) => Void
	method vola(kilometros) { energia = energia - kilometros * 10 }
	
	// XPECT methodType at come --> (Number) => Void
	method come(grms) { energia -= grms * 0.10 }
}
```

And the following xpect tests, and it works.
I will add a new branch for these changes.
But for now we can close this issue.

PR #1151, opened 2017-07-19
Unify double and decimal into number

### LINKED ISSUE #1129: Unify Integer and Decimal classes in a single Number wollok class
Natively we should use only a java.lang.BigDecimal, and show decimal points or commas depending on i18n

fdodino:
Fixes #1129 

Behavior is similar to previous version.


org.uqbar.project.wollok.lib/src/wollok/lang.wlk
```diff
@@ -724,7 +724,8 @@ class List inherits Collection {
 		if (self.isEmpty()) 
 			throw new Exception("Illegal operation 'anyOne' on empty collection")
 		else 
-			return self.get(0.randomUpTo(self.size()))
+			//return self.get(0.randomUpTo(self.size()))
```
npasserini: This changes the behavior of anyOne, without any discussion.
I think we should keep original behavior.

org.uqbar.project.wollok.lib/src/wollok/lang.wlk
```diff
@@ -1055,7 +1128,7 @@ class Number {
 	 * 14.6165.roundUp(3) ==> 14.617
 	 * 5.roundUp(3) ==> 5
 	 */
-	 method roundUp(_decimals)
+	method roundUp(_decimals) native
```
npasserini: Couldn't we just forget the "_" prefix?

org.uqbar.project.wollok.lib/src/wollok/lang.wlk
```diff
@@ -1065,89 +1138,32 @@ class Number {
...
-	override method stringValue() native	
-	
+	 
 	/** 
 	 * Builds a Range between self and end
 	 * 
 	 * Example:
-	 * 		1..4   Answers ==> a new Range object from 1 to 4
+	 * 		1..4      Answers ==> a new Range object from 1 to 4
+	 *      1.0..4.0  Answers ==> a new Range object from 1 to 4
```
npasserini: Given that there is now no difference between 1 and 1.0, I think that this second example only adds confusion.

org.uqbar.project.wollok.lib/src/wollok/lang/Range.xtend
```diff
@@ -39,8 +40,12 @@ class Range extends AbstractJavaWrapper<IntegerRange> {
 	
 	def Integer validate(Object value) {
 		try {
-			return value.wollokToJava(Integer) as Integer
-		} catch (Throwable e) {
+			val castedValue = value.wollokToJava(BigDecimal) as BigDecimal
```
npasserini: I think this cast could be avoided if we add a generic return type to wollokTojava:
`def <T> T wollokToJava(Class<? extends T> type)`, isn't it?

org.uqbar.project.wollok.lib/src/wollok/lang/WNumber.xtend
```diff
@@ -1,63 +1,165 @@
...
+
+	@NativeMessage("<")
+	def lesser(WollokObject other) { wrapped.doubleValue < other.nativeNumber.wrapped.doubleValue }
+
+	@NativeMessage("<=")
+	def lesserOrEquals(WollokObject other) { wrapped.doubleValue <= other.nativeNumber.wrapped.doubleValue }
+
+	def gcd(WollokObject other) {
+		if (!wrapped.isInteger) {
+			throw new IllegalArgumentException("gcd expects receiver to be an integer")
```
npasserini: I would prefer "gcd can only be applied to integer numbers".

Also this should be a pure Wollok exception, but that problem is already reported in this file, so I will create an issue to solve it later.

org.uqbar.project.wollok.launch/.classpath
```diff
@@ -2,11 +2,11 @@
 <classpath>
```
fdodino: !! I don't know why happened this?

org.uqbar.project.wollok.lib/src/wollok/lang.wlk
fdodino: Indeed! I was testing some error and forgot to rollback.

org.uqbar.project.wollok.lib/src/wollok/lang/Range.xtend
fdodino: Will leave this for future change. Now we have

```xtend
def static Object wollokToJava(Object o, Class<?> t) {
````

fdodino:
Well, at least this program compiles ok with constrained type system activated (issue #1146)

image

Well, @npasserini I agree there is no need to introduce right now this functionality, so we can leave this PR open.

fdodino:
We decided with @npasserini  to deprecate this PR favoring next PR with additional features.

fdodino:
Closed PR since I'm now working on a new one.

PR #1162, opened 2017-07-27
Dev 1150 fix delegate constructor to super

### LINKED ISSUE #1150: Confusing message for using super constructor when superclass has none
The following code highlights super() with a message that reads "Invalid super constructor call. It does not exist. Valid constructors are:"... followed by the quickfixes!
When the superclass doesn't have constructors at all, the message shouldn't include that "valid constructors are" part.

```
class A {
	var energia = 0

	method energia() = energia
	method energia(e) { energia = e }
}

class B inherits A {

	var sarasa	
	constructor(e) = super() {
		sarasa = e
	}
	
	method bla(algo) = algo + 1
}
```

Juancete:
Hi! This PR closes #1150. The solution take three different cases when you delegate:

* To a defaut constructor. 
* To a constructor with parameters but
   * Have no definitions at any superclass.
   * Or yo have definitions with other parameters.

And shows different messages. To do this, I had to change the algorithm that lookup for a delegated constructor.


tesonep:
@Juancete could you please resolve the conflict, as I tried, but your code and the existing is modifying the constructor lookup and I don't understand which we want.

Issue #1172, opened 2017-07-28
Constructor for List


nscarcella:
My life would be a little bit brighter if Lists could be created by passing the elements to a constructor. This would save us from creating a ListLiteral node on the WollokJS AST, and (we could use a New Node instead).

I'll make a PR with at possible implementation shortly.

npasserini:
I think there is already a PR about this.

PR #1173, opened 2017-07-28
List constructor for #1172

nscarcella:
See #1172


nscarcella:
@npasserini I saw you where complaining about failing tests. I think this may be the same case because I commented my changes and it still fails.

Let me know if you fix your scenario so I can uncomment this and try again.

tesonep:
@nscarcella the problem you have is a recurrent random error in the tycho tests, it is not related with your code, as your code is not tested! Please add a test checking the new contructor.

nscarcella:
@tesonep Will this do?

nscarcella:
@tesonep I could add a second constructor if that fixes things:

```
constructor(){ super() }
constructor(head,...tail) {
   cons list = this()
   list.add(head)
   list.addAll(tail)
}
```

This might not be necessary on the xtext environment, but it would be very useful in contexts where the list literal is not available (it would be better to parse the literal to a node representing the instantiation of a class than having to add a node specially for this). It has a workaround if this is not possible but, having  such a constructor on the Set, the lack of it on the List feels odd.

tesonep:
Yes, I don't know, Try to add it, I don't know if it works or not. I was just telling why was exploding

fdodino:
Today I made a review.
Unfortunately it has 2 failures in DebugWithoutThreadingTestCase, which has several tests uncomfortable when you change any wollok library file. The solution is activate the log in the debugger for that test and copy the expected state of the debugger, but it is a tricky test.

fdodino:
I'll close this PR and delete this branch, since it will be part of a branch I'm currently working: typeSystem-additionalChecks

PR #1183, opened 2017-07-31
Adding configurable checkings to the validation extensions

tesonep:


org.uqbar.project.wollok.typeSystem/src/org/uqbar/project/wollok/typesystem/TypeSystemWollokValidatorExtension.xtend
```diff
@@ -13,7 +14,7 @@ import org.uqbar.project.wollok.wollokDsl.WFile
  */
 class TypeSystemWollokValidatorExtension implements WollokValidatorExtension {
 
-	
+	@Check(NORMAL)
```
npasserini: I am ok with normal as long as type system is disabled by default.

PR #1184, opened 2017-07-31
Disabling TS :(

tesonep:
I let it here, I will not push the Red (green) button. 
It only applies to new workspaces, old workspaces with the setting in TRUE are not changed.


tesonep:
@npasserini Do you think is needed to change the setting of the old workspaces? For example if somebody is updating or using an old workspace?

npasserini:
I have no idea, @fdodino ?

npasserini:
It is done.

2017-07-31 12:16 GMT-03:00 Pablo Tesone <notifications@github.com>:

PR #1188, opened 2017-08-01
Life ts improvements

npasserini:
Small improvements for code relating the type system with the loaded libraries.

@lgassman Please note that in the future loaded libraries might need to include type annotations, or we need to define a strategy about that.

@tesonep I would like that we have a HO to understand how do the global scopes and caches work. This is crucial for production level type system.


npasserini:
I declined this PR because it has failing tests that I have already solved elsewhere.

PR #1225, opened 2017-08-22
Updated Eclipse VM Arguments

BrianCraig:
The commit changes the default Java VM usage, i had to change this settings on my computer because everything was rendering poorly, and sometimes it would freeze the program.

### DIFF
--- a/org.uqbar.project.wollok.targetplatform/org.uqbar.project.wollok.targetplatform.target
+++ b/org.uqbar.project.wollok.targetplatform/org.uqbar.project.wollok.targetplatform.target
@@ -46,7 +46,8 @@
 <unit id="org.eclipse.egit.mylyn.feature.group" version="4.8.0.201706111038-r"/>
 <unit id="org.eclipse.egit.feature.group" version="4.8.0.201706111038-r"/>
 <unit id="org.eclipse.egit.source.feature.group" version="4.8.0.201706111038-r"/>
-<repository location="http://download.eclipse.org/egit/updates/"/>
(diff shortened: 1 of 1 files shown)


fdodino:
Hi @BrianCraig, could you provide more information about this?

- Which commit?
- How did you change Java VM usage? Where?

Thanks in advance!

npasserini:
@tesonep

fdodino:
I recently installed Wollok and as I've got plenty of memory, I decide to use more memory

```
-XX:PermSize=256M
-XX:MaxPermSize=1024M
```

and it runs perfect.
So maybe we can accept this PR but also recommend to configure memory usage according to every local resources.
What do you think @tesonep ?

tesonep:
Lleno de plata!!!
I don't know. I think the problem is not with the machines having lots of memory, the problem is with the lower end. The difficult question is how low we support. I don't think running Wollok with less of (or even with) 2GB is so nice. 

I don't know how rich are our students 😃 

You are the ones that are in the field, please tell me which is our intended target. Maybe we can start with a lower min and a higher max.

fdodino:
Jajaja, 
here you have a 2016 poll from UTN.
https://docs.google.com/forms/d/1gyOCMSMvQmBb86GNT5YlxElUg42L-Selm9YVO-hG7Ks/edit#responses

I think we could have this configuration : 128 MB - 512 MB for min-max perm size and adjust it if needed.

tesonep:
I ended integrating the change by hand.

fdodino:
:+1:

PR #1252, opened 2017-09-09
Fix #1245 - throwing AssertionException instead of OtherValueExpected…

### LINKED ISSUE #1245: Test expecting exception with wrong message fails red instead of yellow.
image

fdodino:
…Exception

The problem is that yellow bar is set when an AssertionException is thrown.

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok/lib.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok/lib.wlk
@@ -126,7 +126,7 @@ object assert {
 		}
 		catch ex : OtherValueExpectedException 
 		{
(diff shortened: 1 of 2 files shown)


npasserini:
Do we still need OtherValueExpectedException after this change?
Or else... what if we make it a subclass of AssertionException?

fdodino:
I first tried to subclass OtherValueExpectedException from AssertionException.
But 
1) we are not checking against a hierarchy, but just against the fqn, so it didn't help to solve the issue.
2) since we will be deprecating exceptions hierarchy, I decided to do minimal changes.

OtherValueExpectedException is thrown by an internal method in assert, and catched by  3 common methods: throwsException, throwsExceptionLike and throwsExceptionWithMessage.

Issue #1219, opened 2017-08-18
Run all tests from a folder created inside src does nothing


mmatos:
I have a project with the following structure inside of the src folder:
3-objetos-y-mensajes (name of mumuki guide)
    tomYJerry (name of exercise)
      2 wlk files
      2 wtest files
   another folder for other exercise
      1 wlk file
      1 wtest file

When I select the src folder and choose to run all tests, this works, but when I select either of the folders inside of the src folder nothing happens and there is no stacktrace either. The test perspective shows up but it's empty and there is no other feedback.

fdodino:
Yes, since multiple folder manipulation was developed after "Run all test" feature, I haven't tested it with different folder (is a missing functionality)

fdodino:
I could reproduce it: in error log viewer it shows this stack trace

```bash
java.lang.NullPointerException
	at org.uqbar.project.wollok.ui.launch.shortcut.WollokLaunchShortcut.launch(WollokLaunchShortcut.java:55)
	at org.uqbar.project.wollok.ui.tests.shortcut.WollokTestLaunchShortcut.launch(WollokTestLaunchShortcut.java:38)
	at org.uqbar.project.wollok.ui.launch.shortcut.AbstractFileLaunchShortcut._doLaunch(AbstractFileLaunchShortcut.java:62)
	at org.uqbar.project.wollok.ui.launch.shortcut.AbstractFileLaunchShortcut.doLaunch(AbstractFileLaunchShortcut.java:80)
	at org.uqbar.project.wollok.ui.launch.shortcut.AbstractFileLaunchShortcut.launch(AbstractFileLaunchShortcut.java:30)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchShortcutExtension.launch(LaunchShortcutExtension.java:431)
	at org.eclipse.debug.internal.ui.actions.LaunchShortcutAction.runInternal(LaunchShortcutAction.java:85)
	at org.eclipse.debug.internal.ui.actions.LaunchShortcutAction.runWithEvent(LaunchShortcutAction.java:135)
	at org.eclipse.jface.action.ActionContributionItem.handleWidgetSelection(ActionContributionItem.java:565)
	at org.eclipse.jface.action.ActionContributionItem.lambda$4(ActionContributionItem.java:397)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5227)
[...]

Issue #1218, opened 2017-08-18
Run all tests at project level fails when there are spaces in the path


mmatos:
I have a project with the following structure inside of the src folder:
3-objetos-y-mensajes (name of mumuki guide)
    tomYJerry (name of exercise)
       2 wlk files
       2 wtest files
    another folder for other exercise
       1 wlk file
       1 wtest file

When I select the project and choose the option to run all tests in the project, nothing is run and this stacktrace shows up:

java.io.FileNotFoundException: C:\Users\IBM_ADMIN\Documents\UTN\PdeP\Wollok%20PdeP\Wollok%201.6\workspace\ejerciciosMumuki\src\3-objetos-y-mensajes\tomYJerry\tomYElRobot.wtest (The system cannot find the path specified)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(Unknown Source)
	at java.io.FileInputStream.<init>(Unknown Source)
	at org.eclipse.emf.ecore.resource.impl.FileURIHandlerImpl.createInputStream(FileURIHandlerImpl.java:99)
	at org.eclipse.emf.ecore.resource.impl.ExtensibleURIConverterImpl.createInputStream(ExtensibleURIConverterImpl.java:360)
	at org.eclipse.xtext.resource.XtextResourceSet$1.createInputStream(XtextResourceSet.java:298)
	at org.eclipse.emf.ecore.resource.impl.ResourceImpl.load(ResourceImpl.java:1269)
	at org.uqbar.project.wollok.launch.WollokChecker.parse(WollokChecker.java:116)
	at org.uqbar.project.wollok.launch.WollokChecker$1.apply(WollokChecker.java:131)
	at org.uqbar.project.wollok.launch.WollokChecker$1.apply(WollokChecker.java:1)
	at org.eclipse.xtext.xbase.lib.internal.FunctionDelegate.apply(FunctionDelegate.java:42)
	at com.google.common.collect.Lists$TransformingRandomAccessList.get(Lists.java:572)
	at java.util.AbstractList$Itr.next(Unknown Source)
	at org.eclipse.xtext.xbase.lib.IteratorExtensions.fold(IteratorExtensions.java:643)
	at org.eclipse.xtext.xbase.lib.IterableExtensions.fold(IterableExtensions.java:554)
	at org.uqbar.project.wollok.launch.WollokLauncherInterpreterEvaluator.evaluateAll(WollokLauncherInterpreterEvaluator.java:130)
[...]

Issue #1209, opened 2017-08-15
Repeated run configuration


fdodino:
If you run several times your tests, Wollok creates a new configuration file every time:

![image](https://user-images.githubusercontent.com/4549002/29318512-b0264190-81a6-11e7-848a-544bbbf9f53d.png)

I think it would be nice to reuse same run configuration.

fdodino:
Still happening on dev.
Maybe the problem is working copy...

fdodino:
Hi @tesonep , wait wait! :smile: Still happens...

Issue #1189, opened 2017-08-01
WARNINGs in REPL use line and column causing broken link to editor


fdodino:
See this example

```xtend
class A {
    var a
}
```

Run in REPL Console, you see a WARNING line

``` 
[WARNING] /home/fernando/Desktop/IDEs/workspace/tpMusicGuide-2/src/objects.wlk:2:15 WARNING Variable is never assigned
[WARNING] /home/fernando/Desktop/IDEs/workspace/tpMusicGuide-2/src/objects.wlk:2:15 WARNING Unused variable
Wollok interactive console (type "quit" to quit):
>>> 
```

If you click on objects.wlk, link is broken because it is parsing only the line number.
It also happens in wtests files.

PR #1275, opened 2017-09-23
Run all test fixes

fdodino:
Fixes

- #1189 
- #1209 
- #1218 
- #1219 

anim


fdodino:
I'll make an additional change: I want to show in Wollok Test view the name of selected folder if any.

fdodino:
Now it is ok to be reviewed.
Let's say you have this project:

image

If you run from project or src folder, you will see "All tests in project"
image
Every tests has its describe file as a preffix.

If you run from a certain folder (inside src), you see relative path of folder run:
image
You also see tests with associated describe file.

And if you run a single test file, you see  project relative path (instead of full absolute path as before):
image

Issue #1193, opened 2017-08-02
Wollok Game: if you make Gameboard higher or wider, background image is partially applied to cells


fdodino:
Steps to reproduce it:

```xtend
>>> game.height(8)
>>> game.width(8)
>>> game.start()
```

![image](https://user-images.githubusercontent.com/4549002/28868729-b468576c-7750-11e7-8b51-c50b632e7b1f.png)

PalumboN:
We should lock the windows, no?

PalumboN:
Ohh I misunderstood!

PR #1320, opened 2017-12-18
Fix cells creation for resized games

PalumboN:
Fixes https://github.com/uqbar-project/wollok/issues/1193


fdodino:
Well done @PalumboN . 
I fix the conflict, and tested your solution. It works!!

PR #1328, opened 2018-02-21
Avoid to run type system while typing

npasserini:
This configuration changes makes type system run when a file is saved instead of running while the user types. After testing it in a few projects, performance seems reasonable.


npasserini:
Problem persists. Closing.

fdodino:
I tested it, and sounds good to me.

PR #1343, opened 2018-03-08
fix 1341 - allow to annotate constructors types for core classes.

npasserini:
Also we add a first version of constructor expectations (c.f. issue 1342)


fdodino:
Awesome! Really awesome!

The only drawback I've found is that this test is still failing:

```xtend
	// XPECT methodType at m2 --> (Number) => Date
	method m2(day) = new Date(day, day, 2018)
```

This test is located in dates.wlk.xt

npasserini:
Yes, but that one needs for the OpenConstructor strategy we talked about
yesterday.
I will create another issue for that.

2018-03-08 18:02 GMT-03:00 Fernando Dodino <notifications@github.com>:

fdodino:
Perfect. So, shall we merge this PR?

PR #1347, opened 2018-03-09
Fix 1346 constructor call constraints

npasserini:
I understand that this covers all "traditional" constructors, both in the directly reference class and inherited. I used the look up strategy already present (`resolveConstructor`). @fdodino please verify my logic.

We should create another issue / pr for automatic (name-based) initialization which is not considered anywhere in TS.


fdodino:
I changed the base from master to dev-1.6, I don't know if I have to do a rebuild from travis...

npasserini:
@fdodino any comments on this? Do we merge?

fdodino:
Please let me 2 hours and I'll see it...

fdodino:
Perfect! Let's merge it!

Issue #1305, opened 2017-11-03
overriding printString works fine with REPL but not with tests results


marcodicro:
When I override the printString method it works fine when showing an object in the REPL, but when a test fails, the error message doesnt use my printString, it uses the default printString

fdodino:
Hi @marcodicro, could not reproduce it,

![issue1305.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/576a9edc38c9e40e00848c20/125516c66e82c728ace21e0d46e187893687d2fadfbcfd0fe64de29b234324676b5ee17e25d6b42eba447957570e4cf8175d0944fcab8c2cf4bf6d62964e49bc8f3900299f7744eeaa880dbe08b0218e4d84b0b98c4cc9e03f25ef99c2.png)

How did you get this behavior?

marcodicro:
Hi @fdodino, now I realize it has something to do with the printString() for lists.

In my case I was comparing two lists. You can test it in your case like this:

 assert.equals([a],[new B()]) 

It does not use the printString() defined.

Maybe this is the expected behavior.

Thanks!

fdodino:
Right!, this test is failing:

```xtend
	@Test
	def void elementsToString() {
		'''
		object a {
			override method printString() = "es un a"
		}
		
		class B {
			override method printString() = "cierto que era un b!"
		}
		
		program c {
			assert.equals(["es un a", "cierto que era un b!"], [a,new B()].toString())
		}
		'''.interpretPropagatingErrors
	}		
```

I don't know why toString() method isn't redefined, but internalToSmartString() in Collection.

fdodino:
![image](https://user-images.githubusercontent.com/4549002/37793279-ef0e892e-2dec-11e8-9146-7d09d65ae640.png)

marcodicro:
Excellent, Fer! Thanks

fdodino:
:smile: 

But please wait until we merge PR into dev branch.

marcodicro:
Sorry! 😅

Issue #582, opened 2016-01-10
assert.throwsException() error could include the block sourcecode


javierfernandes:
This is the current message that it produces

```java
wollok.lang.Exception: Block should have failed
	at wollok.lib.assert.throwsException(block) [/wollok.wlk:622]
	at  [/testSimplePreConditionForSingleParameterInClass.wtest:12]
```

Which is fin, but could be improved if it also includes the block source code 
Like for this example:

```xtend
test "simplePreConditionForSingleParameterInClass" {
	val b = new Bird()
	assert.throwsException({ => b.fly(0) })
}
```

It should show

```java
wollok.lang.Exception: Block { => b.fly(0) } should have failed
	at wollok.lib.assert.throwsException(block) [/wollok.wlk:622]
	at  [/testSimplePreConditionForSingleParameterInClass.wtest:12]
```

It will probably need to trim new lines and spaces to make if fit. Or even cut it to a limited amount of characters adding "..." in case the block is too long.

Implementation notes:
* As the current method is native this can be done easily. You can access any semantic model object's source code with **org.uqbar.project.xtext.utils.XTextExtensions.sourceCode(obj)**
* In case we code this methods in wollok itself then this requires some kind of **homoiconic** feature for wollok, allowing to treat code as text. Opens up a whole new world.

PR #1360, opened 2018-03-24
Usability issues

fdodino:
**Fixes:**

- #582 
- #1269 
- #1305 
- #1337 
- #1358

### DIFF
--- a/org.uqbar.project.wollok.lib/src/wollok/lang.wlk
+++ b/org.uqbar.project.wollok.lib/src/wollok/lang.wlk
@@ -174,8 +174,6 @@ class Object {
 	 * String representation of Wollok object
 	 */
 	method toString() {
-		// TODO: should be a set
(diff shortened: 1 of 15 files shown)


org.uqbar.project.wollok.lib/src/wollok/lang.wlk
```diff
@@ -174,8 +174,6 @@ class Object {
 	 * String representation of Wollok object
 	 */
 	method toString() {
-		// TODO: should be a set
-		//return self.toSmartString(#{})
```
npasserini: Why is it not a Set?

fdodino: Several months ago I tried to uncomment this line, but I experienced performance issues (tests were very slow, maybe the problem is in Set implementation itself)

PR #1369, opened 2018-04-03
fix: remove unused messages

npasserini:
Actually, if you run tests in the dev env, it produces no warnings, but the prod env does. If searched for the two reported message strings and they do not seem to be used, so I propose to remove them.


fdodino:
PROPERTY_ONLY_ALLOWED_IN_CERTAIN_METHOD_CONTAINERS is the correct message code we should use in message_es.properties, so instead of removing it I renamed it. 

WollokConversion_UNSUPPORTED_CONVERSION_WOLLOK_JAVA is ok to remove it.

I'll push these changes.

PR #1540, opened 2018-10-16
Fix #1491 - validation and quick fix for return expression in parameter

### LINKED ISSUE #1491: Errors in return as an argument - pepita.eat(return 4)
## First problem

This code shows this error:  "Debe retornar un valor en todas las posibles ramas de la ejecución". But the description of the error is incorrect.  It should be something like "there is code that will never be executed"

```Wollok
object foo {
	 method bar(){
	 	pepita.eat(return 4)
	 }
}
```

## Second problem

This error appears in the "Problems" view but not in the editor.

## Third problem

A similar code doesn't show any error:

```Wollok
object foo {
	 method bar(){
	 	return 3.min(return 4)
	 }
}
```

fdodino:
anim

Added new validation and changed validator for "return" in method to avoid a misleading "Must return on every possible flow" message.

Added also a quick fix in order to remove "return" keyword.

I'd like to review the error reported and quick fix messages too.


asanzo:
Love it!

I hadn't seen the oportunity for a quick fix, great feature!!!

PR #1576, opened 2018-12-21
#1564 Fix property validation to let them implement abstract methods

JavierGelatti:
## Problem
Wollok doesn't let abstract methods to be implemented by properties.

For example, with this code:
```
class Arma {
   method artefacto()
}

class EspadaVisual inherits Arma {
      const property artefacto = new Espada()
}
```
doing a `new EspadaVisual()` raises an error, because it doesn't implement the abstract method `artefacto`.

## Solution
Change the logic we use in the validator to take properties into account when determining the non-implemented abstract methods.

### DIFF
--- a/org.uqbar.project.wollok/src/org/uqbar/project/wollok/model/WMethodContainerExtensions.xtend
+++ b/org.uqbar.project.wollok/src/org/uqbar/project/wollok/model/WMethodContainerExtensions.xtend
@@ -134,19 +134,36 @@ class WMethodContainerExtensions extends WollokModelExtensions {
 
 	// rename: should be non-implemented abstract methods
 	def static allAbstractMethods(WMethodContainer c) {
-		val hierarchy = c.linearizeHierarchy
-		val concreteMethods = <WMethodDeclaration>newArrayList
-		hierarchy.reverse.fold(<WMethodDeclaration>newArrayList) [unimplementedMethods, chunk |
-			concreteMethods.addAll(chunk.methods.filter[!abstract])
-			// remove implemented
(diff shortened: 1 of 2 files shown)


fdodino:
@JavierGelatti mergeé en dev el fix para poder compilar xtend en maven, y lo pusheé a tu PR. Nos vemos en febrero! :hugs:

org.uqbar.project.wollok/src/org/uqbar/project/wollok/model/WMethodContainerExtensions.xtend
```diff
@@ -134,20 +134,37 @@ class WMethodContainerExtensions extends WollokModelExtensions {
...
+		abstractMethodContainer.abstractMethods.reject[abstractMethod |
+			abstractMethod.isImplementedBy(concreteMethods, propertyGetters, propertySetters)
+		]
+	}
+	
+	protected def static ArrayList<WMethodDeclaration> allConcreteMethods(WMethodContainer c) {
+		c.linearizeHierarchy.reverse.fold(<WMethodDeclaration>newArrayList) [concreteMethods, methodContainer |
+			concreteMethods.addAll(methodContainer.methods.filter[!abstract])
+			concreteMethods
+		]
```
PalumboN: Este `fold` no es un `flatMap`?

npasserini: Sí, pero me parece que XTend no tiene flatMap, ¿me equivoco?
Igual, si fuera cierto, se puede usar el flatMap de XTendExtensions.

PR #1650, opened 2019-05-27
Fixing assert.equals() when used with one parameter

fdodino:
image

Funny thing is that we had an error in our sanity tests (using `assert.equals` instead of `assert.that`).


fdodino:
FYI @faloi :smile:

org.uqbar.project.wollok.tests/src/org/uqbar/project/wollok/tests/sdk/SetTest.xtend
```diff
@@ -79,7 +79,7 @@ class SetTest extends CollectionTestCase {
 		assert.equals(#{1,2,3}, set.asSet())		
 		const list = set.asList()
 		assert.equals(3, list.size())
-		[1,2,3].forEach{i=>assert.equals(list.contains(i))}
```
npasserini: Ajajajaaj, hermoso, estaba el error en nuestros propios tests.

PR #1670, opened 2019-06-22
Formatter cli #1666

fdodino:
This will allow `wollok-cli` to run formatter externally:

wollokFormat

You can do

```bash
wollok format pepita.wlk // prints formatted file directly to the console
wollok format pepita.wlk -save // saves formatted file directly into pepita.wlk
```


fdodino:
:)

Issue #1605, opened 2019-03-16
Message arguments print null in outline


PalumboN:
In _programs_ and _tests_ we can see the sentences (messages) in the outline. When arguments are values it prints ok, but when they are references, it prints `null`.
![Peek 2019-03-16 16-00](https://user-images.githubusercontent.com/4098184/54480267-035d6700-4805-11e9-8266-b654cd443dd8.gif)

PR #1688, opened 2019-07-11
fix: #1605 - outline message arguments

alete89:
Authors: Mauro Casciati, Alejandro Dini, Nicolás Viotti

image


alete89:
I'll publicly expose myself saying that, honestly, I ran the tests after opening this PR. Which reminded me, aren't there any tests to probe the view, right?

fdodino:
No, we have not, you could try to mock some of Eclipse's behavior, but I think we can rely on visual & manual test for now...

PR #1722, opened 2019-07-30
Dynamic diagram #1710 core objects

fdodino:
newDynamicDiagram

- Several references are joined, separated by ", "
- You don't see global wko references unless you create a local reference in the console


ldevoto:
Ese gif vale más que mil palabras 👏👏👏👏

PR #1895, opened 2020-06-13
Changing splash screen

fdodino:
splash


fdodino:
splash

fdodino:
Last version:
splash

Also added uqbar logo.

